Issue #12, opened 2023-02-17
Zarf package remove fails ungracefully


runyontr:
Lots of issue uninstalling and having hanging resources/namespaces/etc

MxNxPx:
my first experience with it failing was disabling a BB component when packaging/deploying, and the remove failed because it tried to delete the HelmRelease for said component.

```
```

runyontr:
Sounds great to me!

runyontr:
@MxNxPx do you want to make this contribution?

Issue #7, opened 2023-01-04
Update Runtime Security tool to use Neuvector


runyontr:
As of v1.50, Neuvector is available in beta. This should be enabled and tested as part of this deployment

runyontr:
Completed by #15

Issue #6, opened 2023-01-04
Update Zarf package to new BigBang component


runyontr:
this repo should be updated to leverage the new BigBang component in the Zarf.yaml

Blocked by:
- https://github.com/defenseunicorns/zarf/issues/1093

runyontr:
Completed by https://github.com/defenseunicorns/zarf-package-big-bang/pull/15

PR #15, opened 2023-02-23
[WIP] Bigbang component

runyontr:
Proposed change to repo in support of Big Bang Noun.

Blocked by this being released: 

https://github.com/defenseunicorns/zarf/pull/1094

Solves:

#22 
#24 
#7 
#6
#12


cocowow/zarf.yaml
```diff
@@ -0,0 +1,63 @@
...
+        - ../values/kyverno.yaml
+        - ../values/istio.yaml
+        - ../values/loki.yaml
+        - ../values/promtail.yaml
+        - ../values/kiali.yaml
+        - ../values/monitoring.yaml
+        - ../values/neuvector.yaml
+        - ../values/tempo.yaml
+        - ../values/authservice.yaml
+        - ../values/loki.yaml
```
MxNxPx, line 61: duplicate ../values/loki.yaml

defense-unicorns-distro/zarf.yaml
```diff
@@ -0,0 +1,64 @@
...
+        - ../values/kyverno.yaml
+        - ../values/istio.yaml
+        - ../values/loki.yaml
+        - ../values/promtail.yaml
+        - ../values/kiali.yaml
+        - ../values/monitoring.yaml
+        - ../values/neuvector.yaml
+        - ../values/tempo.yaml
+        - ../values/authservice.yaml
+        - ../values/loki.yaml
```
MxNxPx, line 63: duplicate ../values/loki.yaml

runyontr:
I was planning on waiting until zarf has a release with the Big Bang noun.  If we update the docs so people know to use the RC, then we can merge and iterate IMO

values/minio.yaml
```diff
@@ -0,0 +1,66 @@
...
+          mode: PERMISSIVE
+
+  minio:
+    enabled: true
+    values:
+      istio:
+        mtls:
+          mode: PERMISSIVE
+      # https://github.com/minio/operator/blob/master/docs/crd.adoc
+      tenants:
```
blancharda, line 26: I believe this should be `tenant` to match [the BB chart](https://repo1.dso.mil/big-bang/product/packages/minio/-/blob/main/chart/values.yaml#L81)

You may also need to add `podManagementPolicy: "OrderedReady"` in order for the stateful set to come up healthy.

runyontr: I think there's a good conversation aroudn whether MinIO is needed as part of BB.  Perhaps for non cloud Velero backups?

blancharda: Loki can also be configured to use it. 

I think it's also going to be relatively common for 3rd party apps to need a bucket for testing. Might be worth keeping around just for that.

runyontr: In my head, this BB core is what's needed for the platform admins, not for tenants.  If tenants need something we should support adding it, but probably not as part of this distro

blancharda: Makes sense.

I'm not sure the implications of using the default PVC with Loki vs setting it up with a minio bucket.
Doing so might make it easier to transition to actual S3 buckets?

Maybe @UncleGedd, @rjferguson21, @RothAndrew have more informed thoughts?

PR #142, opened 2023-04-24
AWS distro: Updating the rights on the loki folder 

cmharden:
so they can be overlayed when doing a zarf deploy


MxNxPx (approved):
lgtm :rocket: !

PR #154, opened 2023-05-04
Updating defense-unicorns-distro to only delete it's own gitrepos

cmharden:
This pr is to address [this issue](https://github.com/defenseunicorns/zarf-package-big-bang/issues/67)


MxNxPx (approved):
lgtm, nice work @cmharden!

PR #165, opened 2023-05-11
Update package publish pipeline and DUBBD package versions

lucasrod16:
The goal of this PR is to update the publish pipeline to build and publish DUBBD packages and the EKS package

Once this PR is merged, a new `0.0.1` tag will be created/pushed to trigger the publish pipeline

- The EKS package we are using with `eksctl` is not an "official" UDS package, but it would be nice to have a known working package versioned in the registry that we can re-use, rather than rebuilding the package on every pipeline run and leaving open the possibility of ad-hoc changes affecting the functionality.

Use semantic versioning for DUBBD packages


.github/workflows/pipeline.publish.yaml
```diff
@@ -39,30 +35,40 @@ jobs:
...
-        run: zarf package publish zarf-package-*.tar.zst oci://ghcr.io/defenseunicorns/packages
+      - name: Publish DUBBD Package
+        working-directory: defense-unicorns-distro
+        timeout-minutes: 60
+        run: |
+          PACKAGE_NAME="$(yq '.metadata.name' zarf.y*ml)"
+
+          PACKAGE_VERSION="$(yq '.metadata.version' zarf.y*ml)"
+
+          zarf package publish "$PACKAGE_PREFIX-$PACKAGE_NAME-$PACKAGE_ARCH-$PACKAGE_VERSION" "$OCI_URL"
```
runyontr, line 64: I think you need to add the file suffix to get the right file name selected.  There should, however, only be one zarf package in this folder so doing a 

```
zarf package publish *.zst "$OCI_URL

```

should work

defense-unicorns-distro/zarf.yaml
```diff
@@ -1,8 +1,8 @@
 kind: ZarfPackageConfig
 metadata:
-  name: big-bang
+  name: big-bang-distro
   description: "Defense Unicorns Big Bang Distro"
-  version: "###ZARF_PKG_TMPL_BIGBANG_VERSION###"
```
runyontr, line 5: Why the move away from doing this in the config file?

.github/workflows/pipeline.aws.yaml
```diff
@@ -40,14 +38,21 @@ jobs:
         run: zarf package create --confirm
         working-directory: .github/workflows/eks
 
-      - name: Create EKS cluster and Zarf Init
-        run: zarf package deploy zarf-package-distro-eks-multi.tar.zst --confirm --set cluster_name=aws-${GITHUB_SHA:0:7}
-        working-directory: .github/workflows/eks #just holds the zarf-config
+        # We need to create a unique cluster name to avoid naming collisions,
```
runyontr, line 41: Are these changes related to publishing?

defense-unicorns-distro/zarf.yaml
lucasrod16: There wasn't a good reason. Added a `package_version` variable to the config file, and now using it as `###ZARF_PKG_VAR_PACKAGE_VERSION###` in the `zarf.yaml`

.github/workflows/pipeline.aws.yaml
lucasrod16: Made changes back to using commit SHA in cluster names

lucasrod16:
> Now that I look through this, I'm wonder if we want the zarf version to be what's in the Zarf-config file, or what the git tag is?
> 
> This will push the zarf version, regardless of the tag. In some manner I like the tag matching the version of the package, but also like the version being what's in the zarf-config.
> 
> Per haps we should set a command line argument for building to override the version with the GIT_TAG value?

Any particular reason why you would like the zarf version in the config file rather than the git tag?

bdfinst:
> Now that I look through this, I'm wonder if we want the zarf version to be what's in the Zarf-config file, or what the git tag is?
> 
> This will push the zarf version, regardless of the tag. In some manner I like the tag matching the version of the package, but also like the version being what's in the zarf-config.
> 
> Per haps we should set a command line argument for building to override the version with the GIT_TAG value?

If we want the package version in the config file, I think we should create an action that automatically injects it as part of an auto-version flow rather than have people manually enter version numbers.

runyontr:
Right now the version of the zarf package that gets created is tied to a particular value in the zarf-config.yaml that gets checked in to the git-repo.

* This makes the package that gets released, have a version independent of the value of the git tag. This would require coordinate/tools/protection/etc to ensure we don't overwrite a tag b/c someone didn't bump this value

* git tags need to be unique (by some git rules), if we set the value of the zarf package in CI so the zarf package was released with a version matching the git tag, we wouldn't need all that automation and protection

lucasrod16:
> Right now the version of the zarf package that gets created is tied to a particular value in the zarf-config.yaml that gets checked in to the git-repo.

@runyontr 

Yes, currently on the `main` branch, I see that the ['.metada.version' key in our DUBBD `zarf.yaml`](https://github.com/defenseunicorns/zarf-package-big-bang/blob/a462bef89d5cbad4839902b25da6d29d9894485c/defense-unicorns-distro/zarf.yaml#L5) is set to read the value from the `###ZARF_PKG_TMPL_BIGBANG_VERSION###` variable, which is defined in the [`.package.create.set.bigbang_version` in the `zarf-config.yaml`](https://github.com/defenseunicorns/zarf-package-big-bang/blob/a462bef89d5cbad4839902b25da6d29d9894485c/defense-unicorns-distro/zarf-config.yaml#L6).

Based on the decision made in [this ADR](https://github.com/defenseunicorns/product-adrs/blob/main/uds/dubbd/0002-dubbd-versioning.md#decision), I understand that we are aiming to move away from tying DUBBD package versions to the version of Big Bang we are using.
 
>     * This makes the package that gets released, have a version independent of the value of the git tag. This would require coordinate/tools/protection/etc to ensure we don't overwrite a tag b/c someone didn't bump this value
> 
>     * git tags need to be unique (by some git rules), if we set the value of the zarf package in CI so the zarf package was released with a version matching the git tag, we wouldn't need all that automation and protection

Great points. I think setting the package version in CI to the same version as the git tag is a good idea. Once we get an initial `v0.0.1` tag created, a more optimal workflow would be for us to auto-bump tags as well, and set the package version to the new bumped tag in CI. The auto-tag creation/bump won't be in the scope of this PR though

PR #166, opened 2023-05-11
feat: one istio ingress-gw for public and passthru

MxNxPx:
thanks to a major assist by @mjnagel

### DIFF
--- a/values/istio.yaml
+++ b/values/istio.yaml
@@ -1,8 +1,6 @@
 istio:
   enabled: true
   ingressGateways:
-    passthrough-ingressgateway:
-      type: "LoadBalancer"
     public-ingressgateway:
       type: "LoadBalancer"
   values:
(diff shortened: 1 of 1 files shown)


values/istio.yaml
```diff
@@ -25,19 +23,19 @@ istio:
             limits:
               memory: "1Gi"
   gateways:
+    passthrough:
+      ingressGateway: public-ingressgateway
+      hosts:
+        - "keycloak.{{ .Values.domain }}"
```
mjnagel, line 29: Any significance in the change here from `*.{{ .Values.domain }}` to just keycloak? I think it makes sense, but wasn't sure if there would be reasons we'd want to allow other *things* to use the passthrough gateway?

MxNxPx: @mjnagel - good question.  since both the public and passthrough istio gateways would be tied to the same istio ingress gateway, the expectation is that either we define a separate wildcard "hosts" or an explicitly defined list of "hosts" for the passthrough gateway.  as only keycloak is the only (current) need for passthrough, we went ahead and did the explicitly named hosts route.  likely to change down the road though.

MxNxPx:
closing as some of this is likely to be handled by pepr

PR #167, opened 2023-05-11
Moving eks_oidc_provider_arn under the loki terraform

cmharden:


aws/loki/main.tf
```diff
@@ -3,12 +3,21 @@ provider "aws" {
 
 }
 
+data "aws_eks_cluster" "existing" {
```
justin-o12, line 6: nit:  Could this be a variable instead of data?  Like:
```
variable "aws_eks_cluster_name" {
  description = "EKS cluster name"
  type           = "string"
  default       = ""
}
```
and used like:
```
some_item = var.aws_eks_cluster_name
```

aws/loki/main.tf
```diff
@@ -3,12 +3,21 @@ provider "aws" {
...
+  name = var.name
+}
+
+data "aws_caller_identity" "current" {}
 
+locals {
+  oidc_url_without_protocol = substr(data.aws_eks_cluster.existing.identity[0].oidc[0].issuer, 8, -1) # removes "https://"
+  oidc_arn                  = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:oidc-provider/${local.oidc_url_without_protocol}"
```
justin-o12, line 14: Can you parameterize the partition in the ARN?  example: https://registry.terraform.io/providers/hashicorp/aws/4.8.0/docs/data-sources/partition

aws/loki/main.tf
```diff
@@ -38,12 +47,13 @@ output "dynamodb_name" {
...
-variable "name" {
-    description = "Name for cluster"
+output "eks_cluster_oidc_arn" {
+  description = "The ARN of the OIDC Provider of the EKS Cluster"
+  value       = local.oidc_arn
 }
 
-variable "eks_oidc_provider_arn" {
-
+variable "name" {
```
justin-o12, line 55: Related to the above nit, we could replace the `data` and have this be the `eks_cluster_name`.

justin-o12 (changes requested):

aws/loki/main.tf
cmharden: Added better naming for the cluster.

aws/loki/main.tf
cmharden: Good call!!  Implemented.

aws/loki/main.tf
cmharden: Addressed above

justin-o12 (approved):

justin-o12:
> I love the switch to make the API for this zarf package easier to work with.
> 
> My concern about doing this swap now is that we haven't looked at the wholistic view of all the inputs and outputs of each of the UDS packages to understand how we should send data around.
> 
> I think it could make sense to pass the `eks_cluster_name` variable around to various UDS packages to be able to look up what's needed. Do we want to wait for that assessment before changing the interface so we don't have the change it multiple times, or do we take the short term win with this easier change and potentially have to change the interface again?
> 
> My gut says its fine to change it since there's no expectations of a consistent interface, but at some point soon (when the package is "ready") I'd expect interface changes to be a tough decision and one we'd want to avoid if possible.

I agree with taking the short win for now to make this easier.  I understand not wanting to make a wrong decision now, but I don't think it will be difficult to change this interface down the road.  Also, in my mind the EKS cluster could be created in TF right next to this code making it easier to access its properties rather than having to pass around the name and use the data provider.

justin-o12 (approved):

PR #172, opened 2023-05-12
Consuming UDS IaC Modules

justin-o12:
Consuming S3 and KMS modules.


aws/loki/main.tf
```diff
@@ -3,17 +3,39 @@ provider "aws" {
...
 }
 
+# taken from zarf bb repo
+resource "random_id" "default" {
+  byte_length = 2
+}
+
+# taken from zarf bb repo
+locals{
+    kms_key_alias_name_prefix  = "alias/dubbd-aws-testing-${lower(random_id.default.hex)}"
```
runyontr, line 13: Since this module will be used outside of testing, we probably don't want this local variable called "testing"

aws/loki/main.tf
```diff
@@ -3,17 +3,39 @@ provider "aws" {
...
+}
 
+# taken from zarf bb repo
+variable "kms_key_deletion_window" {
+  description = "Waiting period for scheduled KMS Key deletion. Can be 7-30 days."
+  type        = number
+  default     = 7
+}
+
+module "kms_key" {
```
runyontr, line 23: Do we want to be able to allow them to pass in their own KMS key if they already have one made that they want to use?

aws/loki/terraform.tfvars
```diff
@@ -1,4 +1,3 @@
-name = "###ZARF_VAR_NAME###"
+name =  #"###ZARF_VAR_NAME###"
 eks_oidc_provider_arn = "###ZARF_VAR_EKS_OIDC_PROVIDER_ARN###"
-kms_key_arn = "###ZARF_VAR_LOKI_KMS_KEY_ARN###"
-force_destroy = "###ZARF_VAR_LOKI_FORCE_DESTROY###"
\ No newline at end of file
+force_destroy = true #"###ZARF_VAR_LOKI_FORCE_DESTROY###"
```
runyontr, line 3: Should these still be commented out?

aws/loki/main.tf
```diff
@@ -34,9 +56,9 @@ output "s3_bucket" {
   value =  module.S3.s3_bucket
 }
 
-output "dynamodb_name" {
-  value =  module.S3.dynamodb_name
-}
+#output "dynamodb_name" {
```
runyontr, line 59: lets just delete this all together

runyontr (commented):
I would have expected some changes in the Zarf variables in zarf.yaml here since we're not passing a key in anymore

aws/loki/main.tf
justin-o12: Removing "testing".  I will also change this to a variable with a default of the current alias prefix minus the "testing."

aws/loki/main.tf
justin-o12: Ack, removing.

aws/loki/terraform.tfvars
justin-o12: No, fixing.

aws/loki/main.tf
justin-o12: Yeah, that's a good idea.  I'll add in this option in the next revision.

justin-o12:
I tested the latest commit by hardcoding values in the tfvars file instead of the Zarf variables.

wirewc (commented):
@zachariahmiller mentioned in voice, there maybe some other changes that may impact this PR.

aws/loki/main.tf
```diff
@@ -3,60 +3,82 @@ provider "aws" {
...
+  source = "github.com/defenseunicorns/uds-iac-aws-kms"
+  kms_key_description = "KMS Key for DUBBD"
+  kms_key_deletion_window = var.kms_key_deletion_window
+  kms_key_alias_name_prefix = local.kms_key_alias_name_prefix
 }
 
-variable "name" {
-    description = "Name for cluster"
+module "S3" {
+  source = "github.com/defenseunicorns/uds-iac-aws-s3"
```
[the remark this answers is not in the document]
justin-o12: I would like to use a version here, but I don't think we have versioning setup, yet.  What would you suggest here?

aws/loki/main.tf
```diff
@@ -3,60 +3,82 @@ provider "aws" {
...
-
+output "irsa_role" {
+  value = module.S3.irsa_role
 }
 
-variable "key_alias" {
-  description = "alias for KMS Key"
-  default = "bigbang-loki"
+output "s3" {
+  value = module.S3
```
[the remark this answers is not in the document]
justin-o12: The GitHub diff is deceiving.  This [output currently exists in `main`](https://github.com/defenseunicorns/zarf-package-big-bang/blob/main/aws/loki/main.tf#L29).  I'm not sure where any of these outputs are used, and we might be able to get rid of at least some of them.  Outputting the entire S3 module here doesn't seem as helpful as the bucket name or ARN.

Do you know what might be using these outputs?

justin-o12:
> @zachariahmiller mentioned in voice, there maybe some other changes that may impact this PR.

Are these changes that we should discuss or consider now?

zachariahmiller (approved):
LGTM!

MxNxPx (approved):
this was discussed together, and good to go!

cmharden (approved):

aws/loki/main.tf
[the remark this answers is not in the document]
MxNxPx: another PR will follow regarding these source refs

PR #175, opened 2023-05-16
use updated setVariable syntax

rjferguson21:


rjferguson21:
Closing, looks like @kirankt has this fixed over in https://github.com/defenseunicorns/zarf-package-big-bang/tree/cocowow

PR #180, opened 2023-05-17
Add short pre-flight to detect if on an upgrade

mjnagel:
Small pre-flight check to create a zarf var `IS_UPGRADE` set to true/false based on BB HR existence. This var could be used in `onDeploy` actions and/or `onFailure` actions.

Potential funky edge case: the BB HR exists but isn't healthy yet


MxNxPx (approved):
lgtm!

PR #185, opened 2023-05-18
Pull teardown commands into script for reuse

mjnagel:
Teardown commands will likely be reused for `onFailure` on an install, so pulling them out to a separate script file for better future reuse.

Also fixes a bug with the metrics-server deletion step, ensuring it is owned by BB before removing.


MxNxPx (approved):
much cleaner and easier to reuse in other zarf component actions, love it!

PR #186, opened 2023-05-18
fix: update aws zarf loki-module component tf path

MxNxPx:
in the current state, if the machine running zarf package does not have the terraform cli installed, it fails.
fixing the path where it is referencing the terraform stuff for the zarf loki module component.


mjnagel (approved):
LGTM!

runyontr (approved):

PR #188, opened 2023-05-19
Help Zarf catch upgrade failures

mjnagel:
The primary goal of this PR is to have Zarf catch failures of a DUBBD install/upgrade. This is accomplished by altering the `onDeploy.after` actions to add:
- wait on the `bigbang` gitrepository to have the correct BB version
- flux reconciliation of `bigbang` helmrelease (with source)
- kubectl patches to force reconciliation of all BB packages

Secondary addition was to add an `onFailure` action with additional debug (kubectl describe) *if* a helmrelease shows as being in a failed state. If no helmrelease shows as failed the debug log will still contain all helmrelease describes (baked into zarf extension).


MxNxPx (approved):
love it @Michael-Kruggel !

PR #196, opened 2023-05-30
Cleanup values, mTLS STRICT for kiali/promtail

mjnagel:
Couple cleanup pieces:
- random comment
- random newline
- EFK + twistlock disable values (not needed post-2.0)
- logo/image link fix

Functional changes: Removed mTLS permissive override for kiali/promtail. This appears to work fine and has been set as the default in BB for a while now - I believe it was a legacy carryover that isn't necessary in current BB/DUBBD versions.

### DIFF
--- a/aws/values/aws-loki.yaml
+++ b/aws/values/aws-loki.yaml
@@ -1,5 +1,5 @@
 loki:
-  strategy: "scalable" # i think?
+  strategy: "scalable"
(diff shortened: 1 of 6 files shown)


MxNxPx (approved):
@mjnagel - tested this locally and services still work as expected.  lgtm!

PR #201, opened 2023-06-01
Fix handling of flux reconcile behavior

mjnagel:
Changes the `onDeploy.after` behavior to better handle reconciliations by reconciling (or suspend/resuming) in order: gitrepo, helmchart, helmrelease.

Also edits default flux values to have 0 retries, 1min interval, and no rollback for each helmrelease.

Related to https://github.com/defenseunicorns/zarf-package-big-bang/issues/14


mjnagel:
Confirming a few lingering issues with reconcile/resume behavior and ordering then will move to review.

PR #206, opened 2023-06-01
Fix terraform commands in DUBBD-AWS package

lucasrod16:
Closes https://github.com/defenseunicorns/zarf-package-big-bang/issues/205

The way `terraform` commands are currently being executed in the zarf package uses a preexisting binary installed in the $PATH.

This PR updates the `terraform` commands to use the `./terraform` syntax to ensure we are using the zarf-installed terraform binary.

This PR also:

- Updates the name of the job in the `pipeline.dubbd.yaml` workflow from `eksctl` to `clean-install` to be more descriptive of what the job does

- Removes the `pipeline.yaml` workflow. This was being used to catch package creation errors early on, but the same outcome is achieved by moving the package creation steps earlier on in the DUBBD pipelines. This PR moves those package creation steps to be executed early in the DUBBD package pipelines to achieve that.

- Adds step to pipelines that saves zarf debug logs as job artifacts


lucasrod16:
Fixed by https://github.com/defenseunicorns/uds-package-dubbd/pull/347

PR #211, opened 2023-06-02
WIP

UncleGedd:


UncleGedd:
The failed `DUBBD` job is saying the cluster already exists. I'm guessing this is due to a failed cleanup?
```
2023-06-04 21:01:25 [✖]  creating CloudFormation stack "eksctl-8080e27-cluster": operation error CloudFormation: CreateStack, https response error StatusCode: 400, RequestID: 57110b46-4824-4a04-a0c0-b7f6af4eeb8f, AlreadyExistsException: Stack [eksctl-8080e27-cluster] already exists
```

defense-unicorns-distro/zarf.yaml
```diff
@@ -137,7 +137,9 @@ components:
           - cmd: ./zarf tools kubectl wait --for=jsonpath='{.spec.ref.tag}'=###ZARF_PKG_TMPL_BIGBANG_VERSION### gitrepository bigbang -n bigbang --timeout=5m
             description: "Wait for BigBang GitRepository to update to latest version"
             mute: true
-          - cmd: ./flux reconcile helmrelease bigbang -n bigbang --with-source --verbose --timeout 15m
+          - cmd: |
+              ./flux reconcile helmrelease bigbang -n bigbang --with-source --verbose --timeout 15m
+              ./zarf tools kubectl wait --for=condition=Ready --timeout 300s HelmRelease bigbang -n bigbang
```
mjnagel, line 142: Do you have an example CI run where you saw this issue? `flux reconcile` should wait until the HR is fully ready/reconciled so in theory this wait would always pass instantly.

We were seeing some issues where the reconcile errored out in CI that some of us were trying to resolve on Friday - related to the `helmchart` not being ready (different from the `helmrelease`).

UncleGedd: Maybe I'm missing something though. I assumed the above was due to the top-level `bigbang` HelmRelease not being 'Ready' before the `reconcile_helmreleases.sh` script runs

aws/zarf.yaml
```diff
@@ -55,7 +64,8 @@ components:
     actions:
       onCreate:
         before:
-          - cmd: "terraform init -upgrade"
+          - cmd: |
+              terraform init -upgrade -backend-config="bucket=${ZARF_VAR_STATE_BUCKET_NAME}" -backend-config="key=${ZARF_VAR_TF_STATE_KEY}-bucket.tfstate" -backend-config="region=${ZARF_VAR_REGION}"
```
timblaktu, line 68: I'm confused by the usage/flow of variables across these .yaml files. I don't see any definition on this branch of these `ZARF_VAR_*` vars, and I believe they're supposed to use the `###ZARF_VAR_*###` syntax, per [the docs](https://docs.zarf.dev/examples/variables/).

aws/zarf.yaml
```diff
@@ -17,6 +17,15 @@ variables:
 - name: TERRAFORM_VERSION
 - name: TERRAFORM_PROVIDER_AWS_VERSION
 
+- name: TF_STATE_BUCKET_NAME # todo: agree on tf vs terraform for these vars?
```
timblaktu, line 20: IMO The decision on tf var naming convention should be driven by:

1. Make them as short as possible, unless this results in:
    a. human confusion, i.e. "what is this?"
    b. some collision of variable names.
2. Names must work around/with established variables recognized by the tool ecosystem, e.g. [Hashicorp's `TF_*` and `TF_VAR_*` pattern](https://developer.hashicorp.com/terraform/cli/config/environment-variables).

I'd think that to eliminate confusion, we want to differentiate between our terraform-related variables in `zarf` manifests and the official `TF_*` supported by `terraform`. Also, in our selection of what variables to actually use, IMO we should be using as many of the officially-supported `TF_*` vars as possible.

timblaktu (commented):
I'm still trying to understand the usage and flow of variables in the zarf packages, but left some comments/questions.

aws/zarf.yaml
mjnagel: My understanding is that Zarf will add the prefix `TF_VAR_` to every variable, so we would just need `BUCKET_NAME` - https://docs.zarf.dev/docs/create-a-zarf-package/component-actions#creating-dynamic-variables-from-actions - but I may be incorrect and that only works for actions/dynamic variables.

defense-unicorns-distro/zarf.yaml
mjnagel: Resolving - pushed changes to handle this.

aws/zarf.yaml
mjnagel: If I'm not mistaken these need to be `./terraform` to use the bundled terraform binary vs system (not something you introduced but maybe worth fixing in this PR)?

@lucasrod16 was fixing in his PR - https://github.com/defenseunicorns/zarf-package-big-bang/pull/206/files#diff-829a28ee0a6106294551c1fe36af82801a1efaf31daf7b98931881f739883022R81 - but ran into some CI issues I think.

UncleGedd: Yeah it's confusing. My undestanding is that that there's a difference between the Zarf templates (the `###` syntax) and Zarf variables (the `ZARF_VAR_` syntax). See template examples [here](https://docs.zarf.dev/examples/variables/#deploy-time-variables-and-constants) and Zarf variable example [here](https://docs.zarf.dev/examples/component-actions/)

UncleGedd: Also @mjnagel is correct, I need to update that code to be `./terraform`

aws/zarf.yaml
UncleGedd: Nice. I didn't know about the `TF_VAR_` templating, let me test and see if that works

aws/zarf-config.yaml
```diff
@@ -14,6 +14,10 @@ package:
       public_key_file: bigbang.dev.key
       public_cert_file: bigbang.dev.cert
       name: "big-bang-aws" # maybe inject something here for the GH PR?
+      state_bucket_name: uds-dev-state-bucket
+      state_key: tfstate/dev/uds-dev-state-bucket.tfstate
```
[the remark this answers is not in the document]
UncleGedd: Noted! I updated the key for the CI in the latest commit below. As discussed, we can finalize the key and bucket naming convention as a follow on ticket

aws/zarf.yaml
UncleGedd: I take it back! `TF_VAR` is not working as expected, changing back to `ZARF_VAR`

zachariahmiller:
LGTM if you can find a way to push this through... need to update the owners :D

zachariahmiller (approved):

MxNxPx (approved):
lgtm!

Issue #157, opened 2023-05-08
Allow Apps on top of DUBBD to not be from IronBank


runyontr:
When running DUBBD as a runtime environment, there are often applications that need to get deployed that are not from IronBank.  

# Kvyerno Policies

There are current Kyverno policies in place that prevent, by default, installing Deployments, Statefulsets, Job and Pods that come from non-iron bank registries.  

# Options:

## Platform Registry Exception

Create a variable in DUBBD that lets platform admins set any registry that teams can pull from to allow zarf packages with particular registries to be allowed to be installed.

Pros:
* flexible in allowing zarf/non-zarf deployments

Cons:
* requires platform configuration
* Requires improvements to Zarf to allow a variable with a list of strings to be rendered.

## Only monitor Pods

* Let the deployments/Statefulsets/Jobs be free to configure any registry
* Restrict Pods to only be allowed to pull from 127.0.0.1 so they must be installed via Zarf and what's allowed/not is controlled externally

Pros:
* Allows anything to be installed, as long as it was installed with a zarf package

Cons:
* Requires updating the KyvernoPolicies.
* approvals and assurances of the "right"/"safe" things being installed via zarf are removed.  This would require a more robust solution (e.g. signing verification) to provide better checks would be needed.

zachariahmiller:
+1 to platform registry exception option.

ntwkninja:
I think [this](https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/values.yaml?ref_type=heads#L204-223) is much more valuable than which registry you pull from (cryptographic validation of artifacts on admission based upon a provided list of trusted signatories). 

https://kyverno.io/docs/writing-policies/verify-images/

mjnagel:
> I think [this](https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/values.yaml?ref_type=heads#L204-223) is much more valuable than which registry you pull from (cryptographic validation of artifacts on admission based upon a provided list of trusted signatories).
> 
> https://kyverno.io/docs/writing-policies/verify-images/

Would echo this - Big Bang has a policy in place for verifying image are signed from Ironbank (although it isn't enabled/enforced yet). I think it would be valuable to think through how DUBBD/Zarf would handle this policy.

zachariahmiller:
[...]

PR #212, opened 2023-06-02
Added variable for approved registries

runyontr:
Provides variable for #157


cmharden:
Pulled locally and passed testing.

cmharden (dismissed):

mjnagel (dismissed):
LGTM - @cmharden this is probably a good one to document in the var documentation task due to the nuances of how it works. Templating of that list effectively follows this pattern: `<item1>* | <item2>* | <item3>*`, which means if someone wants to just put one registry in that list they should just put `registry.example.com`. If they want to add two registries they would need to do `registry.example.com* | registry.example2.com` in order for things to template out right.

Also worth noting that the default policy will end up having duplicates:
```
        pattern:
          image: registry1.dso.mil* | registry.dso.mil* | registry1.dso.mil* | registry.dso.mil*
            | 127.0.0.1* | 127.0.0.1* | registry1.dso.mil*
```
This appears to work, just an interesting side effect of having a default zarf var + default values.

mjnagel (approved):

PR #235, opened 2023-06-06
fix: switch to pipe

mjnagel:
Revert TF command to use `|`.


mjnagel:
Fixed in https://github.com/defenseunicorns/zarf-package-big-bang/pull/236/files

PR #241, opened 2023-06-07
fix: move teardown back to zarf.yaml

mjnagel:
Having the teardown in a standalone script does not work well since it assumes the script will be available in the local directory when `zarf package remove` runs.

This PR moves all of the teardown commands into a single `cmd` `onRemove.before` action.


MxNxPx (approved):
fine @mjnagel

PR #250, opened 2023-06-07
Use existing Network in CI

zachariahmiller:


MxNxPx:
when we used to have issues with pipelines failing and leaving aws resources stranded, this was going to help with that problem.  now we don't really have that issue anymore, going to close this PR.

PR #252, opened 2023-06-07
fix: set min tls version at gateway

mjnagel:
Adds minimum TLS at the gateway in addition to the mesh config.

See https://repo1.dso.mil/big-bang/product/packages/istio-controlplane/-/issues/78#note_738607

Related to https://github.com/defenseunicorns/zarf-package-big-bang/issues/220


UncleGedd (approved):
LGTM 🚀

Issue #244, opened 2023-06-07
Conditionally Remove KMS and S3 bucket


UncleGedd:
As a delivery engineer
When I remove the DUBBD-AWS package
I want the Loki S3 bucket and  KMS key to be destroyed only if force destroys are set to "true"
So that on longer-lived clusters, removing DUBBD-AWS doesn't impact S3 and KMS

justin-o12:
I thought we could use [Terraform's lifecycle](https://developer.hashicorp.com/terraform/tutorials/state/resource-lifecycle), but I got an error talking about that being reserved for a future use.  I think it's because that isn't compatible with the module we're using at this time.

I spoke with Palassis and Zach, and we agreed that what could work in Rex's use case is a situation where he wants to remove DUBBD but not touch Terraform.  So, I am looking for a way to not perform a `terraform destroy` if the `loki_force_delete` variable is `false`.

There was also talk about generalizing the variable from `loki_force_delete` to something like `ephemeral` to future-proof it.  

I am committing my effort to the `fix-key-arn` branch (at the time, I thought the problem was the ARN, but that's not the case).  The shell script in the `cmd` of the `zarf.yaml` is complaining about:
```
     sh: 1: [: =: unexpected operator
  ✔  Completed "if [ $ZARF_VAR_LOKI_FORCE_DESTROY = "true" ]; then terraf..."
```
After running:
```
zarf p create --confirm && zarf p deploy --confirm ./zarf-package-big-bang-distro-aws-amd64-0.2.1.tar.zst --set name=jno-bigbang-aws-install --set loki_force_destroy=true ; zarf p remove --confirm ./zarf-package-big-bang-distro-aws-amd64-0.2.1.tar.zst --components execute-terraform
```

MxNxPx:
@justin-o12 - I would like to confirm with someone, but i don't know if the deploy var will be available to zarf on destroy like we thought.  new hair-brained idea, what we if create a configmap on deploy and read in from it on destroy...
```yaml
components:
  - name: execute-terraform
    required: true
    actions:
      onDeploy:
        before:
          - cmd: ./zarf tools kubectl create cm -n zarf zarf zarf-bigbang-tf-destroy --from-literal=ephemeral=${ZARF_VAR_EPHEMERAL} -oyaml --dry-run=client | ./zarf tools kubectl apply -f -;
      onRemove:
        before:
          - cmd: |
[...]

PR #266, opened 2023-06-11
Delete both S3 and KMS when force_delete is set to true

UncleGedd:
Attempt at resolving #244 

~~2 paths tested:~~
~~- When force_destroy is set to 'true' both S3 bucket and KMS key are deleted~~
~~- When force_destroy is set to 'false" only the S3 bucket is deleted~~

### DIFF
--- a/aws/zarf.yaml
+++ b/aws/zarf.yaml
@@ -100,7 +100,8 @@ components:
             dir: run/loki
       onRemove:
         before:
-          - cmd: "terraform destroy -auto-approve"
+          - cmd: |
+              if [ "$(terraform output force_destroy)" = true ]; then terraform destroy -auto-approve; fi
             dir: run/loki
   - name: outputs
     required: true
--- a/aws/zarf.yaml
+++ b/aws/zarf.yaml
@@ -101,7 +101,7 @@ components:
       onRemove:
         before:
           - cmd: |
-              if [ "$(terraform output force_destroy)" = true ]; then terraform destroy -auto-approve; fi
+              if [ -d "run/loki" ] && [ "$(terraform output force_destroy)" = true ]; then terraform destroy -auto-approve; fi
             dir: run/loki
   - name: outputs
     required: true
--- a/aws/zarf.yaml
+++ b/aws/zarf.yaml
@@ -101,8 +101,16 @@ components:
       onRemove:
         before:
           - cmd: |
-              if [ -d "run/loki" ] && [ "$(terraform output force_destroy)" = true ]; then terraform destroy -auto-approve; fi
-            dir: run/loki
+              if [ -d "run/loki" ]; then
+                cd run/loki
+                if [ "$(terraform output force_destroy)" = true ]; then
+                  terraform destroy -auto-approve
+                else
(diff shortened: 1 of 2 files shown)


UncleGedd:
Probably needs tests for both paths. I'm thinking we can use this as opportunity to refactor out the IaC specific code into another folder and run more "unit-style" tests only on it

justin-o12:
I _think_ the intended behavior is:
```
if force_delete == true:
  # remove all of terraform infra
  remove kms
  remove s3
else:
  # don't remove any terraform infra
```
Therefore, we don't have to target the s3 module if `force_delete == false`.

UncleGedd:
Spoke with @brianrexrode this morning, you are correct @justin-o12 . Plus, the IaC that is untouched should be able to be reused

justin-o12:
> 

To reuse it, sounds like we might need to parameterize the S3 bucket like the KMS key to either reuse or create a new one depending on the input.  Is that right?

UncleGedd:
> To reuse it, sounds like we might need to parameterize the S3 bucket like the KMS key to either reuse or create a new one depending on the input. Is that right?

I think we have it covered. If the same backend config is used, it should just use the existing state (which includes the S3 bucket and KMS key).

Just tested this manually:
- when `force_destroy` is "false"
    - the bucket and key are not destroyed during `zarf package remove`
    - after performing `zarf package remove` and then re-deploying, the existing S3 bucket and KMS key are used

justin-o12 (approved):
LGTM!

aws/zarf.yaml
```diff
@@ -100,7 +100,8 @@ components:
             dir: run/loki
       onRemove:
         before:
-          - cmd: "terraform destroy -auto-approve"
+          - cmd: |
+              if [ "$(terraform output force_destroy)" = true ]; then terraform destroy -auto-approve; fi
             dir: run/loki
```
MxNxPx, line 105: @UncleGedd - going to test tomorrow morning, but I am not sure I understand how this would work if someone did a `zarf package deploy` on one machine (pipeline) and then on another machine tried a `zarf package remove` if that would work as desired.  you probably tested that scenario, but just in case you didn't i am hoping to try tomorrow.  also, please ignore this feedback if that is just a prereq for folks to have a copy of the `run/loki` directory locally.

UncleGedd: Great point! @zachariahmiller brought up the same point. I think our `zarf package remove` currently assumes that the `.terraform` directory is still available. That said, we should probably handle that ASAP. Maybe for now it's safer to go with your solution of putting the vars in a ConfigMap (including all the TF state configs) and hack around the lack of variables during the remove

UncleGedd: As discussed, I'll create another ticket to address the `onRemove` and testing deficiencies

justin-o12 (approved):
LGTM

wirewc:
I really like that latest check to see if the `run/loki` folder.

mjnagel (approved):

MxNxPx (approved):
well done, approved!

PR #272, opened 2023-06-12
fix: Add AWS LB annotations

mjnagel:
Fixes missing annotations. Added only to dubbd-aws since these would not be applicable for k3d, etc.


MxNxPx (approved):
lgtm!

UncleGedd (approved):
LGTM (man we need some more tests)

lucasrod16 (approved):

Issue #221, opened 2023-06-05
Enable tetrate istio distribution


ntwkninja:
As a user required to meet fips compliance, I need the ability to utilize the tetrate istio distribution in DUBBD. I can currently override these values in my zarf package but would like this feature to be supported.

zachariahmiller:
picking up this late requirement for release next week. Appreciate Gabe raising this up, but as a product team we need to get better at identifying things like this earlier.

mjnagel:
```yaml
istio:
  enterprise: true
```
https://repo1.dso.mil/big-bang/bigbang/-/blob/master/chart/values.yaml#L171-173

This should be a quick change to make, but pending the discussion around whether this should be set by default and/or exposed as a zarf var.

zachariahmiller:
waiting for decision from @austenbryan before implementing this.

austenbryan:
Recommend we use Tetrate's Istio distro @zachariahmiller and that we remove this issue as being blocked.

Rationale from Tetrate PM on contract with Platform One

"Yes it does include keeping them updated.
Our renewals occur yearly with them in the March - April timeframe. We’re currently on year 3. Next renewal will be year 4. We’re not worried about them not renewing. But even if that’s ever the case. It’s in our best interest to continue to keep the images in iron bank up to date. Hope this helps.""

PR #275, opened 2023-06-13
Switch istio to TID by default

mjnagel:
Switches istio "enterprise" mode on by default (Tetrate Istio Distribution). Everything came up healthy and all ingress worked fine in a quick test.

Closes https://github.com/defenseunicorns/uds-package-dubbd/issues/221


zachariahmiller (approved):
LGTM!

Issue #277, opened 2023-06-13
Refactor IaC in Zarf pkg and add tests


UncleGedd:
Currently, whenever we test IaC specific things in DUBBD-AWS, we have to spin up an entire EKS cluster which creates a long feedback loop for devs. This issue to migrate some of the IaC-specific components out of the main DUBBD-AWS `zarf.yaml` and into the `aws/loki` folder, as well as, write a "unit" test for IaC-specific components

UncleGedd:
Blocked for now until we decouple S3 and IRSA, need to create an IRSA TF module

wirewc:
@UncleGedd proposed solution for decoupling S3 and IRSA in a single repository. 

If we move all the IRSA functionality into a submodule in the same repository, we may get the modular like behavior that is desired.  

Presently, the attempt perform a dual purpose (IRSA and non-IRSA) behavior will fail as Terraform will always need the IRSA variables to be populated even if they are not used. If a submodule is used, the variables should be able to be safely ignored.

austenbryan:
No longer in scope given UDS CLI and IaC Tool work

PR #278, opened 2023-06-13
Refactors DUBBD-AWS IaC and introduces testing

UncleGedd:
Addresses: https://github.com/defenseunicorns/uds-package-dubbd/issues/277

- refactors the Loki IaC into it's own `zarf.yaml` and imports its components into the DUBBD-AWS `zarf.yaml` so we can now test the Loki IaC in isolation
- adds Terratest to `aws/loki`
- adds workflow for Terratest and only runs when files in the `aw/loki` directory are modified
- makes EKS e2e tests run nightly or on modifications to certain files


aws/loki/main.tf
```diff
@@ -26,20 +26,21 @@ locals {
   oidc_arn                  = "arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:oidc-provider/${local.oidc_url_without_protocol}"
   generate_kms_key          = var.create_kms_key == true ? 1 : 0
   kms_key_arn               = var.kms_key_arn == null ? module.generate_kms[0].kms_key_arn : var.kms_key_arn
+  create_irsa               = var.create_irsa == true ? true : false
```
justin-o12, line 29: Is this used?  It looks like you reference the `var` below in the S3 module.

aws/loki/variables.tf
```diff
@@ -13,6 +13,11 @@ variable "key_alias" {
   default     = "bigbang-loki"
 }
 
+variable "create_irsa" {
+  description = "Create IRSA roles and plicy"
```
justin-o12, line 17: nit: typo

bdfinst (dismissed):
LGTM

.github/workflows/test-loki-bucket.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Upload test log
+        uses: actions/upload-artifact@v2
+        if: always()
+        with:
+          name: test-log
+          path: /tmp/gotest.log
+          if-no-files-found: error
+
+  test-loki-bucket-zarf:
```
lucasrod16, line 48: We may want to add a dependency on passing tests for the zarf package/deployment steps to run:

```yaml
    needs: test-loki-bucket-tf
```

.github/workflows/test-loki-bucket.yaml
```diff
@@ -0,0 +1,82 @@
...
+        uses: defenseunicorns/setup-zarf@main
+        with:
+          # renovate: datasource=github-tags depName=defenseunicorns/zarf versioning=semver
+          version: v0.27.1
+          download-init-package: ${{ inputs.download-init-package }}
+
+      - name: Build Loki Zarf Package
+        working-directory: aws/loki
+        run: |
+          zarf package create
```
lucasrod16, line 67: We will need to add the `--confirm` flag here

.github/workflows/test-loki-bucket.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Deploy Loki Zarf Package
+        working-directory: aws/loki
+        run: |
+          zarf package deploy zarf-package-*.tar.zst \
```
lucasrod16, line 72: We will need to add the `--confirm` flag here

lucasrod16 (changes requested):

.github/workflows/test-loki-bucket.yaml
UncleGedd: done!

aws/loki/variables.tf
UncleGedd: done!

aws/loki/main.tf
UncleGedd: Fixed, using locals in that S3 module block now

.github/workflows/test-loki-bucket.yaml
UncleGedd: love it, adding

aws/loki/zarf.yaml
```diff
@@ -0,0 +1,108 @@
...
+    description: "Name of the DynamoDB table used for Terraform state locking"
+  - name: REGION
+    description: "The AWS region to run the Terraform in"
+
+  - name: LOKI_FORCE_DESTROY
+    description: "Provide a means to force destroy the Loki storage (S3) bucket even if it contains data"
+    default: "false"
+
+components:
+  - name: download-terraform
```
justin-o12, line 32: We had discussed moving this back to `aws/zarf.yaml` so other, future IaC components do not have to include this.  If we don't do it in this PR, we can refactor this as we add the other components (e.g. EKS).

justin-o12 (approved):
All LGTM.  We can refactor where we download Terraform if we decide it's necessary.

lucasrod16 (commented):
@UncleGedd should we be implementing tests and zarf packaging for IaC modules in this repo?

It seems like we might benefit from doing those things in the source IaC repo, and only consume the published zarf package by importing the OCI package into our dubbd-aws package.

Thoughts?

UncleGedd:
> @UncleGedd should we be implementing tests and zarf packaging for IaC modules in this repo?
> 
> It seems like we might benefit from doing those things in the source IaC repo, and only consume the published zarf package by importing the OCI package into our dubbd-aws package.
> 
> Thoughts?

Yeah I see where you're coming from. A few thoughts:
1. OCI skeletons [don't work as expected](https://github.com/defenseunicorns/zarf/issues/1808) for components involving files
2. The terratest tests are IMO akin to a "unit test." In this case it's testing that our DUBBD-AWS-specific Terraform configs are working as expected (ie. the the Terraform code in `aws/loki`)
3. I don't think we should necessarily put a `zarf.yaml` in the TF modules, at least not othis one. IMO the `zarf.yaml` that we import from `aws/loki` should contain opinionated DUBBD configuration, as opposed to the less opinionated configus exposed in the TF modules

Def open to more thoughts/feedback on this though!

aws/loki/zarf.yaml
UncleGedd: moved!

justin-o12 (dismissed):
LGTM

mjnagel (changes requested):
CI is failing on `terraform output -raw irsa_role` in the AWS package (`Output "irsa_role" not found`) - I was trying to track down where/why and I think its related to moving `create_irsa` to a local?

lucasrod16:
> > @UncleGedd should we be implementing tests and zarf packaging for IaC modules in this repo?
> > It seems like we might benefit from doing those things in the source IaC repo, and only consume the published zarf package by importing the OCI package into our dubbd-aws package.
> > Thoughts?
> 
> Yeah I see where you're coming from. A few thoughts:
> 
>     1. OCI skeletons [don't work as expected](https://github.com/defenseunicorns/zarf/issues/1808) for components involving files
> 
>     2. The terratest tests  in `aws/loki` are IMO akin to a "unit test." In this case it's testing that our DUBBD-AWS-specific Terraform configs are working as expected (ie. the the Terraform code in `aws/loki`)
> 
>     3. I don't think we should necessarily put a `zarf.yaml` in the TF modules, at least not this one. IMO the `zarf.yaml` that we import from `aws/loki` should contain opinionated DUBBD configuration, as opposed to the less opinionated configus exposed in the TF modules
> 
> 
> Def open to more thoughts/feedback on this though!

Thanks! Sounds good to me

zachariahmiller:
@UncleGedd isn't it outputting null for the irsa role because you are passing in `--set create_irsa=false` in the zarf command? That would make sense to me. Regardless, this is maybe a little complicated for what you are actually aiming to accomplish.  I'm in support of making a generic irsa role anyways. Also, how would you feel about adding the IaC tests you are trying to do here as an example/loki tests/loki inside the s3 bucket module or to be created irsa module?

UncleGedd:
> Regardless, this is maybe a little complicated for what you are actually aiming to accomplish.

Exactly this, the amount of workarounds I'm having to do to make this work feel off. 

 > adding the IaC tests you are trying to do here as an example/loki tests/loki
 
 I did add another test to the S3 module [here](https://github.com/defenseunicorns/terraform-aws-uds-s3/blob/98e72f23e1914b01012f751a213f7eda3c693cff/test/e2e/examples_test.go#L94), but if we refactor the IRSA into another module then I think this entire [commit](https://github.com/defenseunicorns/terraform-aws-uds-s3/commit/39a10229a7b2b4f2529d1ae1a72cec9b8a999f6e) could be reverted

UncleGedd:
> outputting null for the irsa role because you are passing in --set create_irsa=false in the zarf command?

Yes, I want it to:
- default to true
- but be false when testing and running the `aws/loki` CI
- but be true when running the e2e tests with the cluster

Again, WAY too complicated, all to get around some AWS components that are tightly coupled

zachariahmiller:
> > outputting null for the irsa role because you are passing in --set create_irsa=false in the zarf command?
> 
> Yes, I want it to:
> - default to true
> - but be false when testing and running the `aws/loki` CI
> - but be true when running the e2e tests with the cluster
> 
> Again, WAY too complicated, all to get around some AWS components that are tightly coupled and could be separated

Ah okay I think some of that context was lost on the mobile GitHub view. So yeah that would seem to require multiple zarf.configs to handle appropriately without adding multiple inputs and more testing logic. 

Only comment on the separate irsa module is that delivery is still consuming the s3 module that creates the irsa role so the implementation would need to still exist in the s3 bucket (call irsa module behind true/false) or would need to be a breaking change. I think this can be accomplished without issue and then just change the loki module to add irsa at the same level as the s3 bucket instead of nested inside, which _should_ facilitate what you are trying to accomplish. Still, if the tf output for the role name/arn is needed for the full deployment how do you plan on remedying that without getting into the same conditional rabbit hole?

zachariahmiller:
But doesn't it still need to be conditionally called in the loki module (based on full cluster or not full cluster test)? And as a product of that be conditionally producing an output (which granted you could control to be like "" if it wasn't present) 

I'm all for what you're suggesting if you think you have a way around the current issue. Maybe tomorrow you can show me what you mean real quick and it will just click. In the meantime, I'm good with this immediate change in direction.

UncleGedd:
man this is old!

PR #279, opened 2023-06-13
119 document local dev env setup and uds iac quick start as an onboarding resource

timblaktu:
This PR integrates a new `k3d-local` `zarf` package under the `k3d/` folder of `uds-package-dubbd`, and updates documentation for the `k3d` workflow, automating/dogfooding as much as possible with `zarf` tooling to maximize learning during onboarding.

Closes defense-unicorns/uds-package-dubbd#119.


k3d/README.md
```diff
@@ -1,51 +1,101 @@
...
 ```bash
-# Login to the registry
 set +o history
 export REGISTRY1_USERNAME="YOUR-USERNAME-HERE"
 export REGISTRY1_PASSWORD="YOUR-PASSWORD-HERE"
 echo $REGISTRY1_PASSWORD | zarf tools registry login registry1.dso.mil --username $REGISTRY1_USERNAME --password-stdin
 set -o history
+```
+
+### [Install kubectl](https://kubernetes.io/docs/tasks/tools/#kubectlhttps://kubernetes.io/docs/tasks/tools/#kubectl)
```
Racer159, line 39: `zarf tools kubectl` could be used (but raw kubectl can be nice too)

k3d/local/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Deployment
+
+```bash
+zarf package deploy --confirm zarf-package-k3d-local-<ARCH>-<ZARF_VERSION>.tar.zst 
+```
+
+e.g. `zarf package deploy --confirm zarf-package-k3d-local-amd64-v0.27.0.tar.zst`
+
+[^1]: Noting that only global `zarf` options may be declared, and that `--confirm` is not a global option.
```
Racer159, line 22: This isn't entirely true - sub items (like `package.create.output_directory`) can be configured too, but `--confirm` is explicitly excluded to encourage interactive deployments and reviews of packages (if it were hidden in a `zarf-config` file less experienced users may not get the option to review a deployment to see if they are deploying (or removing!) the correct package.

k3d/local/zarf.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - source: scripts/k3d.sh
+        target: scripts/k3d.sh
+        executable: true
+    actions:
+      onDeploy:
+        after:
+          - cmd: './scripts/k3d.sh ###ZARF_PKG_TMPL_K3D_CLUSTER_NAME###'
+      onRemove:
+        after:
+          - cmd: 'k3d cluster delete ###ZARF_PKG_TMPL_K3D_CLUSTER_NAME###'
```
Racer159, line 38: It may be worth noting somewhere that unlike many Zarf packages this doesn't actually work in an air gap and is thus for online dev only

k3d/local/README.md
timblaktu: Gotcha, thanks @Racer159. Then [this page](https://docs.zarf.dev/docs/the-zarf-cli/#using-a-config-file-to-make-cli-command-flags-declarative) needs to be corrected accordingly, no? It says:

> Users can configure the zarf init, zarf package create, zarf package deploy command flags, as well as any global flags, through a config file.

k3d/README.md
timblaktu: @Racer159 thanks for heads-up on this. Are the `zarf tools` like `kubectl` statically embedded in the zarf binary? If so, how does zarf decide which version to use? Sorry if this is documented somewhere, I couldn't find it. [The cli command docs](https://docs.zarf.dev/docs/the-zarf-cli/cli-commands/zarf_tools_kubectl) imply "i just works" without saying _how_.

..but in general, I want to leverage any tools embedded in zarf to minimize dependencies and config/setup steps. Thanks again for the hint.

timblaktu: @wayne i see in [the code](https://github.com/defenseunicorns/zarf/blob/1aeae1b584ec36cd3ce252358c8ccbeabbce0dfe/src/cmd/tools/kubectl.go#L11) that zarf imports the upstream k8s cli code, so i think i answered the question about embedding, only thing remaining is figuring out the relationship between a zarf version and k8s version.

k3d/local/zarf.yaml
timblaktu: @racer159 to work in an airgap, this would need to include the container registry init component, and push the k3s images to it for k3d to use, right? Anything else?

k3d/values/local-dev.yaml
```diff
@@ -1,10 +1,13 @@
+# override neuvector values required for k3d
 neuvector:
   enabled: true
   values:
     containerd:
       enabled: false
     k3s:
       enabled: true
+    monitor:
+      install: false
```
MxNxPx, line 10: is there a reason this is being set to false?

k3d/zarf.yaml
```diff
@@ -2,7 +2,7 @@
 kind: ZarfPackageConfig
 metadata:
   name: big-bang-distro-k3d
-  description: "Defense Unicorns Big Bang Distro"
+  description: "Defense Unicorns Big Bang Distro K3D Environment"
   # x-release-please-start-version
   version: "0.2.1"
   # x-release-please-end
```
MxNxPx, line 8: this would need to be added to the https://github.com/defenseunicorns/uds-package-dubbd/blob/main/release-please-config.json to be versioned alongside the other packages

mjnagel (dismissed):
Can you rebase this + resolve conflicts to make sure things are up to date with `main`?

k3d/local/zarf-config.yaml
```diff
@@ -0,0 +1,15 @@
...
+#   - https://docs.zarf.dev/docs/the-zarf-cli/#using-a-config-file-to-make-cli-command-flags-declarative
+#   - https://docs.zarf.dev/examples/variables/#create-time-package-configuration-templates
+log-level: info
+architecture: amd64
+package:
+  create: 
+    max_package_size: 1000000000
+    set:
+      # TODO: enforce/check `which zarf` is this version
+      zarf_version: v0.27.0
```
MxNxPx, line 11: renovate should auto PR newer zarf versions if you add a comment right above the zarf_version line like this - https://github.com/defenseunicorns/uds-package-dubbd/blob/main/.github/actions/create-bb-package/action.yaml#L36

k3d/values/local-dev.yaml
timblaktu: @MxNxPx not sure, I took these lines from a Slack thread a few weeks ago. I believe they came from @mjnagel ?

k3d/zarf.yaml
timblaktu: @MxNxPx I believe this is now in place after merging main. Can you confirm?

timblaktu:
> Can you rebase this + resolve conflicts to make sure things are up to date with `main`?

@MxNxPx I didn't rebase, but I did merge main, resolving conflicts around the time you sent this. Is rebasing necessary since we're squashing commits on merge?

k3d/README.md
```diff
@@ -1,51 +1,103 @@
...
+  2. `zarf tools registry login ghcr.io -u USERNAME`
+
+#### `registry1.dso.mil` AuthN 
 
 ```bash
-# Login to the registry
 set +o history
 export REGISTRY1_USERNAME="YOUR-USERNAME-HERE"
-export REGISTRY1_PASSWORD="YOUR-PASSWORD-HERE"
+export REGISTRY2_PASSWORD="YOUR-PASSWORD-HERE"
```
MxNxPx, line 34: typo here perhaps?

k3d/values/local-dev.yaml
MxNxPx: got it - then i think we would want to remove those lines

k3d/local/zarf-config.yaml
timblaktu: I see, thanks. This reminds me that all my testing was done using the zarf commands directly, so I need to go back and integrate/test a new k3d-local make target. 

What I need here, regardless of whether zarf is run directly or in renovate container, is to pick up the _actual_ version of the zarf process running. Do you know how to fetch that from the env? I figured it would exist as a `PKG_TMPL_*` kind of like `ARCH` is but haven't been able to find it yet.

mjnagel:
> I didn't rebase, but I did merge main, resolving conflicts around the time you sent this. Is rebasing necessary since we're squashing commits on merge?

@timblaktu as long as main is merged & conflicts resolved we should be totally fine - no need for a true rebase :)

k3d/zarf.yaml
MxNxPx: i was looking at the wrong file also - my bad!

k3d/local/zarf-config.yaml
timblaktu: @MxNxPx that renovate comment is in a GH action workflow. Where would I put that in my case, i.e. running all this via make?

MxNxPx: @timblaktu - renovate itself will cut a PR to update that version based on the comment in any yaml.  although the `v` prefix` in the version would need to be removed in order to have renovate regex to match it.

mjnagel: I might just be confused here - when you say you are running this via make...was that just for local testing or did you intend to push up a makefile?

k3d/values/local-dev.yaml
timblaktu: These lines have been removed in latest commit. Thanks!

k3d/local/zarf-config.yaml
timblaktu: @MxNxPx I'm referring to the top-level Makefile in uds-package-dubbd. There aren't any make targets for k3d-local yet (to my knowledge), and what I'm saying is that up till now I've NOT been using make to test k3d-local, instead just calling zarf directly to create/deploy/remove it. As such, I'm not exercising renovate at all WRT k3d-local.

I'm planning to simply wrap these zarf calls in new make targets in the top-level Makefile in uds-package-dubbd, so that we leverage renovate's ability to handle tool versioning.

My TODO that started this thread is there bc currently this k3d-local package needs to know what version of zarf is being used to create the package, bc it downloads the corresponding init package and refers to it by filename which includes version. Since the zarf_version here is tracked as a create-time `PKG_TMPL` zarf template, essentially a hard-coded constant in the zarf-config.yaml, there's nothing to ensure this version is "correct" for a given build environment.

I had assumed the easiest thing to do would be to automatically set the version template value using some available zarf environment variable, but I can't find support for this anywhere. (kind of like how you can fetch the current ARCH of the current package using `###ZARF_PKG_ARCH###`..)

I see now at least 2 other solutions:

1. `make k3d-local` can simply write `zarf version` into `zarf-config.yaml` on `zarf package create`
2. do away with the handling and need-to-know of zarf version altogether, since I'm now downloading the init package using `zarf tools download-init`, which:
    * automatically fetches the correctly-versioned init pkg, and 
    * accepts `-o init_pkg_filename` arg, which means I can just refer to the file as versionless `init` at both `create` and `deploy`-time.

2 is simpler, unless we actually need to know the version for some other reason. I'm guessing not, and that if so, I should be able to wrap that need in a zarf call. `zarf` seems designed to always to the right thing, version-wise.

mjnagel: Oh, I honestly did not realize there was a makefile in this repo - it looks like it hasn't been touched in a while and honestly could probably just be removed in favor of our docs that guide people through using their own `zarf` binary and specific commands to create/deploy. I don't think the make targets are currently referenced anywhere/used in CI.

For what its worth I like option 2 you mentioned - it keeps it simple and removes the need for version.

timblaktu: @palassis furthermore, in the bigger picture we need to remember that this k3d-local package is for dubbd development. As such, in addition to ensuring the deployed init pkg == the current zarf version, it needs to *enforce* that this is identical to the zarf version being used by the top-level dubbd package.

To that end, I believe 2 above is the way to go - remove any dependency/usage/consciousness of the zarf version being used.

k3d/README.md
timblaktu: Good catch, I copied this from the main k3d README and probably vim-botched it somehow. ;-) @MxNxPx how do you feel about me replacing these instructions for container reg authN with the way I did it above in the GitHub AuthN section?

k3d/local/zarf-config.yaml
timblaktu: @mjnagel thanks. I also got confirmation from @lucasrod16 that the Makefile wasn't being used in CI. I also see that we are not using the build harness in CI. Is that you're understanding?

Yes, I plan to implement 2 above.

timblaktu: @MxNxPx @mjnagel @lucasrod16 my latest commits have removed the Makefile and "zarf-version-consciousness" from the k3d-local package..

k3d/local/README.md
timblaktu: Tracked here: https://github.com/defenseunicorns/zarf/issues/1833

k3d/README.md
timblaktu: This was resolved in commits last week which consolidated docs about authN to container registries.

wirewc (approved):

k3d/README.md
```diff
@@ -1,51 +1,111 @@
...
-- K3D cluster has zarf init package deployed (with git-server component)
-- MetalLB deployed to support multiple LoadBalancer service types
-- Working kube context (`kubectl get nodes` <-- this command works)
+### [Install zarf](https://docs.zarf.dev/docs/getting-started/#installing-zarf)
+
+### [Install docker](https://docs.docker.com/install/https://docs.docker.com/install/)
+
+### Gain Access to GitHub Container Registry (`ghcr.io`)
+
+1. Login to your GitHub Account with access to the `defenseunicorns` organization.
```
mjnagel, line 17: nit: `defenseunicorns` org access isn't required is it? All the images/artifacts pulled here should be public to anyone with a github account?

k3d/README.md
```diff
@@ -1,51 +1,111 @@
...
-# Run the zarf package deploy command
+# Deploy the zarf package, EITHER the..
+#   locally-created package .zst file
 zarf package deploy --confirm zarf-package-dubbd-*.tar.zst
-
-# (Alternatively) Deploy from OCI
-# Login to the registry
-# Run the zarf package deploy command with the desired DUBBD OCI package reference 
+#   OR the published OCI package 
 zarf package deploy oci://ghcr.io/defenseunicorns/packages/dubbd-k3d:0.0.1-amd64 --oci-concurrency=15
```
mjnagel, line 105: Might be worth updating this comment/command to reflect the latest tag and/or link to release page to get the latest.

timblaktu:
> Added comments - I think this approach is OK, though again Make may be better for local development since while Zarf can be a shell wrapper, Make was designed for it.

Agreed, this was an odd application of `zarf` that was done by a curious newbie. If the UDS team decides to use `make` again,  this "zarfification" of  k3d-local pkg could easily be moved to a recipe there.

k3d/README.md
timblaktu: @mjnagel correct, ghcr isn't required to deploy the k3d cluster. These are considered consolidated instructions for deploying dubbd on local k3d cluster, and that's why ghcr is mentioned here.

timblaktu: @mjnagel I get your point now... :/ Yes, I suppose the defenseunicorns org is irrelevant here. I'll remove.

k3d/local/zarf.yaml
```diff
@@ -0,0 +1,78 @@
...
+        target: init.tar.zst
+    actions:
+      onCreate:
+        before:
+          - cmd: |
+              F="downloads/zarf-init-###ZARF_PKG_ARCH###-$(zarf version).tar.zst";
+              if [ -f "$F" ]; then
+                echo "Using cached zarf init package: $F..."
+              else
+                (set -x; zarf tools download-init -o downloads)
```
mjnagel, line 46: `zarf tools download-init` will always download based on the system architecture, which might not match the hardcoded package architecture (amd64). I was trying to use `-a amd64` which should be a global flag but it doesn't seem to be respected.

In my case (on an ARM mac) I was seeing this error:
```
  ✔  Downloaded https://github.com/defenseunicorns/zarf/releases/download/v0.27.1/zarf-init-arm64-v0.27.1.tar.zst                                                                                                                         
     cp: downloads/zarf-init-amd64-v0.27.1.tar.zst: No such file or directory                                                                                                                                                             
     ls: *.zst: No such file or directory                                                                                                                                                                                                 
     downloads:                                                                                                                                                                                                                           
     total 1048704                                                                                                                                                                                                                        
     -rw-r--r--  1 mjnagel  staff   497M Jun 21 08:37 zarf-init-arm64-v0.27.1.tar.zst                                                                                                                                                     
     drwx------  3 mjnagel  staff    96B Jun 21 08:36 .                                                                                                                                                                                   
     drwxr-xr-x  8 mjnagel  staff   256B Jun 21 08:36 ..                                                                                                                                                                                  
     ERROR:  Failed to create package: unable to add component: unable to run component before action: command                                                                                                                            
             "F="downloads/zarf-init-amd64-$(zarf version).tar.zst";; i..." failed after 0 retries
```

mjnagel (changes requested):

k3d/README.md
timblaktu: Done.

k3d/local/zarf.yaml
timblaktu: @mjnagel arch is declaratively hard-coded in zarf-config.yaml to amd64, this is why the instructions don't include `-a`. I suppose the action item here is to update the README to indicate this only works on amd64.

mjnagel: If we want to document as a limitation I'm fine with that. In general I would expect that the package should be able to be built on any architecture, then only deployed on `amd64` (DUBBD works this way). In this case that is not possible because init will be downloaded based on the system architecture.

k3d/README.md
```diff
@@ -1,51 +1,124 @@
...
+
+The `k3d-local` package itself may only be deployed to `amd64`, a limitation inherited from DUBBD via IronBank. This is declaratively hard-coded in [`zarf-config.yaml`](./zarf-config.yaml).
+
+Typically, zarf package _creation_ is supported on any arch (analogous to [cross-compilation](https://en.wikipedia.org/wiki/Cross_compiler)). The `k3d-local` package can only be _created_ on amd64 systems, because:
+
+1. the `k3d-local` package downloads the `zarf init` package using `zarf tools download-init`
+1. `zarf tools download-init` does not respect `-a`/`--architecture` CLI args or `architecture:` configuration declarations.
+    * This is a [zarf bug, tracked here](https://github.com/defenseunicorns/zarf/issues/1837).
+
+Currently, Package creation is currently only supported on amd64 systems, due to a bug that `zarf tools download-init` ignores -a/--architecture` args, and always uses the system architecture._
```
mjnagel, line 79:
```suggestion
```
I think this is duplicative of the above?

mjnagel (approved):
LGTM otherwise - thanks for all the work on this @timblaktu 🥳

mjnagel (approved):

mjnagel (dismissed):

timblaktu:
> LGTM otherwise - thanks for all the work on this @timblaktu partying_face

@mjnagel @wirewc would you mind approving? The PR is still blocked requiring review.

wirewc:
@timblaktu I'm looking at something quickly before approving.

wirewc (approved):
:+1:

Issue #14, opened 2023-02-21
Test Updates of BB Versions for DUBBD


runyontr:
For testing we should be sure we can rollback on failures

Deploy 1.52
Deploy 1.53
Rollback to 1.52

cmharden:
Test upgrade of dubbd with a failed state
* document behavior
* Expected state: Flux is still working and zarf just shows failed to deploy

Decision?
What should the rollback proc be for now?
* Redeploy of the previous version?
* Zarf needs to handle(on deploy fail action?)
* Some other option?

mjnagel:
Big piece of nailing this down is having zarf properly catch errors so that we can use `onFailure` actions. https://github.com/defenseunicorns/zarf-package-big-bang/pull/188 puts *part* of that into place by triggering reconciliation of all packages on updates, and waiting on the Big Bang gitrepo/helmrelease to update to the version of BB expected. This works well for upgrades that include a BB version update but does not necessarily account for other scenarios perfectly. 

We may want to consider pulling in the `flux` CLI to directly trigger a `flux reconcile hr bigbang -n bigbang --with-source` rather than approaching with annotations and homegrown waits.

mjnagel:
https://github.com/defenseunicorns/zarf-package-big-bang/pull/195 is a second small piece, "fully rollback" (aka uninstall) when an installation (non-upgrade) fails to deploy. This is the simpler scenario where we can delete all of DUBBD. One decision point - do we leave Flux always, delete it always, or try to determine if Flux "pre-existed" and handle uniquely based off of that? DUBBD itself assumes we are deploying Flux (not pre-existing) so I lean towards delete always, but not sure if that hinders extensibility.

mjnagel:
> We may want to consider pulling in the `flux` CLI to directly trigger a `flux reconcile hr bigbang -n bigbang --with-source` rather than approaching with annotations and homegrown waits.

Tested this yesterday and the `flux` CLI brings a lot of benefits - it has caught 100% of failures in my tests so far and is still relatively small to pull in the binaries. In combo with the checksums on all HRs (coming in BB 2.3.0 hopefully - https://repo1.dso.mil/big-bang/bigbang/-/commit/f3adb3ca06c8324f6632594f461f610e5f61ac89) upgrades should be triggered close to instantly after `zarf package deploy` which should ensure that the `onDeploy.after` waits will catch the state post-upgrade.

mjnagel:
[...]

PR #280, opened 2023-06-13
DUBBD-AWS Upgrade Documentation

mjnagel:
Adds upgrade documentation for DUBBD-AWS. Also fixes a few spots with the wrong package name.

Preview rendered: https://github.com/defenseunicorns/uds-package-dubbd/blob/aws-docs/aws/README.md

Closes https://github.com/defenseunicorns/uds-package-dubbd/issues/14

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Pre-built Zarf Package of [DoD-Platform-One/big-bang](https://github.com/DoD-Pla
 
 ## Prerequisites
 
-- Zarf is installed locally with a minimum version of [v0.26.1](https://github.com/defenseunicorns/zarf/releases/tag/v0.26.1)
+- Zarf is installed locally with a minimum version of [v0.27.1](https://github.com/defenseunicorns/zarf/releases/tag/v0.27.1)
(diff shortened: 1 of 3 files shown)


cmharden (dismissed):

UncleGedd:
Re: rollbacks. Should we recommend Option 3? Does that not jack with the Zarf state in the cluster which could hurt future upgrades?

mjnagel:
> Re: rollbacks. Should we recommend Option 3? Does that not jack with the Zarf state in the cluster which could hurt future upgrades?

@UncleGedd - it shouldn't mess with "zarf state" (the secret) since that doesn't get updated until the component succeeds if I'm not mistaken. Even if it did, it shouldn't cause any technical challenges for future upgrades outside of a confusing current state (ex: DUBBD 1.2.3 deployed but secret says 1.2.4).

I don't have a strong leaning either way, although we were considering automating the `helm rollback` path as an option in DUBBD which I think is part of why we included it here.

UncleGedd (approved):
Thanks for this! LGTM

PR #285, opened 2023-06-14
ci: config temp set releases to patch only

MxNxPx:
temporary change so all versions are patches (for now)

manual override option: still can force a version with a commit message:
```sh
git commit --allow-empty -m "chore: release 0.3.0" -m "Release-As: 0.3.0"
```

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -4,8 +4,7 @@
       "release-type": "simple",
       "draft": false,
       "changelog-path": "CHANGELOG.md",
-      "bump-minor-pre-major": "true",
-      "bump-patch-for-minor-pre-major": "true",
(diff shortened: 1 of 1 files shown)


mjnagel (approved):
LGTM - just to confirm we still intend to release minors occasionally but would trigger that manually rather than have release-please decide for us? Do we have any idea on when we might want minors?

MxNxPx:
> LGTM - just to confirm we still intend to release minors occasionally but would trigger that manually rather than have release-please decide for us? Do we have any idea on when we might want minors?

For the time being, yes - we would only do minors manually.  Also for now, we want to hold off on a minor update until Friday at earliest.

Issue #297, opened 2023-06-15
Expose config for public/private Load Balancers


UncleGedd:
As a delivery engineer
I want the ability to deploy public or private tenant LBs
So that my DUBBD-AWS package is adaptable to my environment

As part of this issue, we also need to automate the provisioning and deprovisioning of of public vs private LBs during upgrades

Issue #71, opened 2023-04-04
Identify how to access private ELBs


runyontr:
We need an easy to use way for non-technical people to access private VirtualServices

Related to #29 

* tailscale/headscale
* wireguard
* SSM
* https://github.com/slackhq/nebula
* https://www.boundaryproject.io/

wirewc:
@oates and @UncleGedd, this issue fell off the old project board. I'm bumping this for consideration of priorities as this would be helpful in terms of DUBBD adoption.

MxNxPx:
Handy (likely biased) comparisons:
https://tailscale.com/compare/

PR #299, opened 2023-06-16
feat: allow configuring public/private tenant LB

mjnagel:
Changes:
- Adds a `private_tenant_lb` zarf var to configure whether the provisioned tenant elastic loadbalancer is public or private
- Adds actions before/after to cycle the tenant loadbalancer service if this value changes on an upgrade (ensures a new ELB is provisioned)

Notes:
- all logic was added to the AWS package as it is currently AWS specific but we could move to "core" DUBBD in the future if other flavors require the same variable/functionality
- bash scripting is not intended to be the final solution for "upgrade logic" - there are other proposed routes of doing things in a more programmatic approach w/testing, etc but as there is not a clear decision on that approach yet this was done with basic scripting
- this basic pattern would likely apply to all future upgrade logic as well - `onDeploy.before` actions to check before state, then a final component to take action based on said state

Closes https://github.com/defenseunicorns/uds-package-dubbd/issues/297

### DIFF
--- a/aws/values/aws-istio.yaml
+++ b/aws/values/aws-istio.yaml
@@ -3,7 +3,7 @@ istio:
     admin-ingressgateway:
       kubernetesResourceSpec:
         serviceAnnotations:
-          service.beta.kubernetes.io/aws-load-balancer-internal: "###ZARF_VAR_PRIVATE_TENANT_LB###"
+          service.beta.kubernetes.io/aws-load-balancer-internal: "true"
           service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: "true"
     tenant-ingressgateway:
       kubernetesResourceSpec:
--- a/aws/values/aws-istio.yaml
+++ b/aws/values/aws-istio.yaml
@@ -3,10 +3,10 @@ istio:
     admin-ingressgateway:
       kubernetesResourceSpec:
         serviceAnnotations:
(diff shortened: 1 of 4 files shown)


aws/values/aws-istio.yaml
```diff
@@ -3,10 +3,10 @@ istio:
     admin-ingressgateway:
       kubernetesResourceSpec:
         serviceAnnotations:
-          service.beta.kubernetes.io/aws-load-balancer-internal: "true"
+          service.beta.kubernetes.io/aws-load-balancer-internal: "###ZARF_VAR_PRIVATE_TENANT_LB###"
```
runyontr, line 6: shouldn't this be a different variable?  i.e. `###ZARF_VAR_PRIVATE_ADMIN_LB###`?

mjnagel: Oof - definitely messed this up. I intended to leave this one as hardcoded `"true"`, and not allow configuration on it but open to other's opinions. Thanks for catching - if we want it configurable I'll make it a separate var and ensure the upgrade "job" will handle config changes as well.

mjnagel:
Noting that this does not provide docs/configuration of external access - that will be done as a part of https://github.com/defenseunicorns/uds-package-dubbd/issues/71

MxNxPx (approved):
lgtm :rocket:

PR #301, opened 2023-06-20
docs: adding verbiage around release expectations

MxNxPx:


README.md
```diff
@@ -58,7 +58,15 @@ set -o history
...
 ```
 
-## Creating releases
+## Releases
+
+### When to expect new releases
+
+New DUBBD versions can be expected (at a minimum) shortly after each Big Bang release is available.
+
+DUBBD versions will be released as fixes or features are introduced. Following semantic versioning for releases, a new release version is created upon implementing changes that are categorized as fixes or features. For example, if there is a new Big Bang release pulled into DUBBD - that would be considered a new feature, and the minor version number would be incremented in the corresponding DUBBD release.
```
mjnagel, line 67: nit: I would think if there was a BB patch release with a bug fix, it's probably also going to be classified as a bug fix on our end too. Maybe worth rewording this example to...
```suggestion
DUBBD versions will be released as fixes or features are introduced. Following semantic versioning for releases, a new release version is created upon implementing changes that are categorized as fixes or features. For example, if there is a new Big Bang feature release pulled into DUBBD - that would be considered a new feature, and the minor version number would be incremented in the corresponding DUBBD release.
```

mjnagel (approved):

README.md
mjnagel: Maybe a broader convo about how we version based on BB updates - we probably have two options?
1. Dive into each release and see how it affects DUBBD - is it just fixes, new package updates/features, etc? Can DUBBD do any upgrade actions to handle breaking changes, etc?
2. "Match" BB versions - patch from BB = patch for DUBBD, minor for BB = minor for DUBBD, etc.

I think option 1 is better to more accurately reflect semver, and forces us to identify if there's areas that DUBBD can handle upgrades more smoothly than upstream BB.

MxNxPx: removed the sentence altogether to reduce confusion

mjnagel (approved):

Issue #216, opened 2023-06-04
Create a faster feedback loop for DUBBD in CI


UncleGedd:
As a UDS dev
I want a more lightweight cluster for DUBBD in CI
So that I don't have to wait for EKS clusters to be created/destroyed
```[tasklist]
### Tasks
- [ ] https://github.com/defenseunicorns/uds-package-dubbd/issues/337
```

zachariahmiller:
I have a PR into the branch lucas is working on integrating with renovate https://github.com/defenseunicorns/uds-aws-ci-k3d/pull/8

Once those are both merged in we can cut a tag and begin on the PR into the dubbd repo.

UncleGedd:
Completed

PR #313, opened 2023-06-22
ci: Add workflows for k3d package

lucasrod16:
Relates to https://github.com/defenseunicorns/uds-package-dubbd/issues/216

- Adds a pipeline that deploys the DUBBD k3d package on a k3d cluster
- Establishes a pattern for spinning up a k3d cluster in AWS for testing all DUBBD packages

Once this PR is merged, the plan is to make the necessary changes to be able to transition testing of the DUBBD AWS package to k3d.


wirewc:
I don't see the value in having a remote workflow when it could be in single file only. But this comment will not without my own approval.

wirewc (approved):

lucasrod16:
> I don't see the value in having a remote workflow when it could be in single file only. But this comment will not without my own approval.

@wirewc good point. My reason for doing it this way was to stick with the existing pattern we are using for testing the dubbd-aws package. A [reusable workflow is defined that runs all of the deployment/test steps](https://github.com/defenseunicorns/uds-package-dubbd/blob/main/.github/workflows/test-aws-package.yaml), and a [calling workflow](https://github.com/defenseunicorns/uds-package-dubbd/blob/main/.github/workflows/ci-on-pull-request.yaml) is used to implement the reusable test workflow and is what triggers the tests to run on every PR commit.

One benefit to the existing pattern we are using is that it establishes a clear separation of concerns. One workflow is used to implement the deploy/test logic, while another workflow calls that test workflow and is only concerned about orchestrating the triggering events and path filters. One helpful way to think about this is to think as the workflows themselves as functions. You can think of the reusable test workflow as a function that is scoped to perform the very specific operations of deploying and testing our packages, and the calling workflow as a function that is scoped to orchestrate the triggering events and path filters for the test function.

We can certainly look at ways to improve this pattern if it's something we decide we should refactor. But it would be scope creep in this PR to refactor the existing pattern we are using.

PR #317, opened 2023-06-22
docs: add uds core diagram

MxNxPx:


mjnagel (commented):
Should we add the docs folder to our ignored workflow paths? 👀 

https://github.com/defenseunicorns/uds-package-dubbd/blob/docs/.github/workflows/ci-on-pull-request.yaml#L8

docs/README.md
```diff
@@ -0,0 +1,5 @@
+# docs
+
+## UDS Core Diagram
```
mjnagel, line 3: Should this be called `DUBBD AWS Diagram`? UDS Core diagram colors would look a little different (i.e. EKS provided by UDS Core rather than user).

Even if we left this as UDS Core, probably should add AWS

.github/workflows/ci-on-pull-request-aws.yaml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
     paths-ignore:
+      - 'docs/**'
```
lucasrod16, line 8: What do you think about adding these patterns to `paths-ignore` rather than `docs/**`?

```yaml
- "**.png"
- "**.svg
```

Those along with the existing `**.md` pattern should ignore all files that would be included in the `docs/` directory, and any other future locations as well.

docs/README.md
MxNxPx: changed to DUBBD AWS

.github/workflows/ci-on-pull-request-aws.yaml
MxNxPx: agree - updated the exclude paths to the ones you suggested @lucasrod16

lucasrod16 (dismissed):
lgtm!

mjnagel (dismissed):
:shipit:

PR #320, opened 2023-06-22
fix: Update workflow path filters

lucasrod16:
Relates to https://github.com/defenseunicorns/uds-package-dubbd/issues/216

Adds file paths in the workflows that are ignored, so that when there are changes to them, the workflows are not triggered

This creates a faster feedback loop by not running the pipelines for changes that do not impact the package or the pipelines themselves. For example, if I make a change to only the k3d package, only the k3d pipeline will run, and the AWS/EKS pipeline will be skipped because there are no changes to the AWS package or pipelines.

### DIFF
--- a/.github/workflows/ci-on-pull-request-aws.yaml
+++ b/.github/workflows/ci-on-pull-request-aws.yaml
@@ -4,11 +4,14 @@ on:
   pull_request:
     branches:
       - main
-    paths:
-      - ".github/workflows/ci-on-pull-request-aws.y*ml"
-      - "aws/**"
-      - "defense-unicorns-distro/**"
-      - "values/**"
+    paths-ignore:
+      - ".gitignore"
+      - "LICENSE"
+      - "**.md"
+      - "**.json"
+      - "k3d/**"
+      - ".github/workflows/ci-on-pull-request-k3d.y*ml"
+      - ".github/workflows/test-k3d-package.y*ml"
   workflow_call:
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-on-pull-request-k3d.yaml
```diff
@@ -7,6 +7,7 @@ on:
     paths:
       - ".github/workflows/ci-on-pull-request-k3d.y*ml"
       - "defense-unicorns-distro/**"
+      - "values/**"
       - "k3d/values/**"
       - "k3d/zarf-config.y*ml"
       - "k3d/zarf.y*ml"
```
mjnagel, line 13: could just make this...
```suggestion
      - "k3d/**"
```
to match the aws pattern

.github/workflows/ci-on-pull-request-aws.yaml
```diff
@@ -4,8 +4,11 @@ on:
   pull_request:
     branches:
       - main
-    paths-ignore:
-      - '**/README.md'
+    paths:
+      - ".github/workflows/ci-on-pull-request-aws.y*ml"
```
mjnagel, line 8: This doesn't fully capture everything that could change to effect the workflow - wondering if we should also capture the `.github/workflows/test-aws-*.yaml` files?

mjnagel (approved):
LGTM :shipit:

.github/workflows/ci-on-pull-request-aws.yaml
lucasrod16: As discussed in a conversation in slack, it would be more simple to maintain if we use the `paths-ignore` keyword instead of `paths`

PR #322, opened 2023-06-22
Integrate EKS Module

justin-o12:
Integrate EKS module into DUBBD-AWS.


.github/workflows/ci-on-pr-eks.yaml
```diff
@@ -0,0 +1,54 @@
...
+    paths:
+      - "aws/eks/**"
+      - "!*.md"
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  test-loki-bucket-tf:
```
mjnagel, line 17: Should be renamed to reflect EKS

.github/workflows/ci-on-pr-eks.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Install Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: 'aws/loki/test/go.mod'
+
+      - name: Set up gotestfmt
+        uses: gotesttools/gotestfmt-action@v2
+
+      - name: Run Loki Tests
```
mjnagel, line 42: Same comment here - rename to reflect EKS.

aws/uds-core-aws/zarf-config.yaml
```diff
@@ -21,7 +21,7 @@ package:
       # -- Name of existing Terraform state bucket
       state_bucket_name: uds-dev-state-bucket
       # -- Key path to Terraform state file within the bucket
-      state_key: tfstate/dev/uds-dev-state-bucket.tfstate
+      state_key: tfstate/dev/uds-dev-2-state-bucket.tfstate
```
justin-o12, line 24: leftover testing artifact

aws/uds-core-aws/zarf.yaml
```diff
@@ -135,6 +167,22 @@ components:
             dir: run/loki
             setVariables:
               - name: LOKI_S3_ROLE_ARN
+  - name: zarf-init
+    required: true
+    actions:
+      onCreate:
+        before:
+          - cmd: wget -nc https://github.com/defenseunicorns/zarf/releases/download/v0.27.1/zarf-init-amd64-v0.27.1.tar.zst
```
justin-o12, line 175: Should we attempt to parameterize the Zarf version?  If so, I can do this and move the `ZARF_VERSION` variable out of dubbd-aws and into this `zarf.yaml`.

aws/dubbd-aws/zarf.yaml
```diff
@@ -19,6 +19,9 @@ variables:
 - name: TERRAFORM_VERSION
 - name: TERRAFORM_PROVIDER_AWS_VERSION
 
+- name: ZARF_VERSION
+  default: "0.27.1"
+
```
mjnagel, line 24: Not seeing anywhere this var is used in this package - is this meant for the uds-core-aws package?

aws/uds-core-aws/zarf.yaml
UncleGedd: I don't think we should parameterize the Zarf version unless we have some sort of testing matrix in CI indicating that the packages can be deployed on multiple versions of Zarf

UncleGedd: For example, we know that there are issues with Zarf v0.27.0. If we parameterize, it makes it seem like users can bring whatever version of Zarf they like

mjnagel, line 175: Should the version be a zarf var so that it is more easily carried around/reused? Also could wire up with Renovate if done that way.

mjnagel (commented):
A few hopefully small comments, otherwise LGTM.

aws/uds-core-aws/zarf.yaml
mjnagel: Resolving this to combine with above discussion.

mjnagel: I like making this a var:
- reuse the version arg to make it easier to change in the future
- renovate config is easier if we parameterize
- similar to the pattern with flux, BB, tf, etc - we don't really want the user to override those/they can't but they are variables to simplify the above two

.github/workflows/test-uds-core-aws-install.yaml
```diff
@@ -29,67 +29,54 @@ jobs:
...
-          echo ${{ steps.get_cluster_name.outputs.eks_cluster_name }}
-
-          zarf package deploy zarf-package-*.tar.zst \
+          zarf package deploy zarf-package-uds-core-aws-*.tar.zst \
             --set loki_force_destroy=true \
-            --set name=${{ steps.get_cluster_name.outputs.eks_cluster_name }} \
+            --set vpc_id=vpc-03c1cd73aece8bfea \
+            --set vpc_cidr=192.168.0.0/16 \
+            --set private_subnets=subnet-065bdf771fa18e8be,subnet-053b6dab6cf6276ea,subnet-0f381300221588640 \
+            --set public_subnets=subnet-0de30cd58a6a6d763,subnet-041516854dd66437f,subnet-0283cdc0228b519ea \
```
mjnagel, line 43: With this being hardcoded...will we have problems running multiple CI runs concurrently? Or is multiple EKS clusters in the same VPC 🆗 ?

justin-o12: I think this should be ok with the SHA used in the cluster name and backend key.

UncleGedd: Multiple EKS clusters in the same VPC should be fine, can always adjust CIDR ranges

aws/uds-core-aws/zarf.yaml
UncleGedd: Good point! I was picturing it as a deploy time var, yeah I agree with making in a create time var

.github/workflows/ci-on-pr-eks.yaml
justin-o12: Addressed.

.github/workflows/ci-on-pr-eks.yaml
justin-o12: Addressed.

aws/dubbd-aws/zarf.yaml
justin-o12: Moved this variable and now using it for UDS core.

aws/uds-core-aws/zarf.yaml
justin-o12: Now using a template variable for the Zarf version.

MxNxPx: @justin-o12 - as long as you do something similar to [this](https://github.com/defenseunicorns/uds-package-dubbd/blob/main/.github/actions/create-zarf-package/action.yaml#L36-L37), then renovate should pick it up and keep up with new versions

aws/eks/zarf.yaml
```diff
@@ -0,0 +1,83 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/defenseunicorns/zarf/main/zarf.schema.json
+kind: ZarfPackageConfig
+metadata:
+  name: dubbd-aws-eks-iac
+  description: "Defense Unicorns Big Bang Distro"
+  # x-release-please-start-version
+  version: "0.2.3"
```
MxNxPx, line 7: this needs to be updated to match the number in the [release-please json](https://github.com/defenseunicorns/uds-package-dubbd/blob/main/.release-please-manifest.json).  sadly if those don't match, it doesn't matter that it is in between the comments.

UncleGedd (approved):

PR #329, opened 2023-06-23
Adds UDS Core AWS Package

UncleGedd:
- refactors repo structure for both `dubbd-aws` and `uds-core-aws`
- renames a lot of files/fields that referred to a "BB package" that doesn't exist


.github/workflows/test-uds-core-aws-upgrade.yaml
```diff
@@ -0,0 +1,118 @@
...
+          working-dir: aws/uds-core-aws
+          download-init-package: true        
+          timeout-minutes: 60
+          
+      - name: Set eks_cluster_name
+        id: get_cluster_name
+        env:
+          SHA: ${{ github.sha }}
+        run: |
+          echo "eks_cluster_name=uds-cre-aws-upgrade-${SHA:0:7}" >> $GITHUB_OUTPUT
```
mjnagel, line 48: Was `cre` over `core` a conscious choice based on length limits...or just a typo?

UncleGedd: typo lol

UncleGedd: but it was nice of you to assume it was intentional 😆

aws/uds-core-aws/zarf.yaml
```diff
@@ -0,0 +1,183 @@
...
+- name: STATE_BUCKET_NAME
+  description: "Name of the pre-existing Terraform state S3 bucket"
+- name: STATE_KEY
+  description: "Path to the state file key in the state bucket"
+- name: STATE_DYNAMODB_TABLE_NAME
+  description: "Name of the DynamoDB table used for Terraform state locking"
+- name: REGION
+  description: "The AWS region to run the Terraform in"
+
+- name: PRIVATE_TENANT_LB
```
mjnagel, line 31: This is missing the new `PRIVATE_ADMIN_LB` and associated actions.

justin-o12 (approved):
LGTM

aws/uds-core-aws/zarf.yaml
UncleGedd: 👍 adding; I think I copied over the code before doing the rebase

mjnagel (commented):
Will need to update the path for release-please to match the new one for DUBBD: https://github.com/defenseunicorns/uds-package-dubbd/blob/adds-uds-core-aws/release-please-config.json#L9

mjnagel (commented):
Will need to update the release workflow with the new dir for dubbd-aws: https://github.com/defenseunicorns/uds-package-dubbd/blob/adds-uds-core-aws/.github/workflows/tag-and-release.yaml#L28

UncleGedd:
Noting that I expect the `CI UDS Core AWS / test-upgrade` job to fail because it hasn't been pushed to GHCR

aws/uds-core-aws/zarf.yaml
```diff
@@ -0,0 +1,183 @@
...
+  description: "Provision a private/internal load balancer for the tenant ingress gateway"
+  default: "true"
+
+- name: LOKI_FORCE_DESTROY
+  description: "Provide a means to force destroy the Loki storage (S3) bucket even if it contains data"
+  default: "false"
+
+components:
```
mjnagel, line 39: Might be missing something here, but shouldn't we just be importing the components from `../dubbd-aws`? Rather than copy-paste all the definition for each component? Ex:

```
  - name: download-terraform
    required: true
    import:
      path: ../dubbd-aws
```

mjnagel: (and ideal future state this might be importing from the skeleton, or something like that?)

UncleGedd: Yeah we should absolutely be doing that, I did the copy past for now because @justin-o12 is working on this [PR](https://github.com/defenseunicorns/uds-package-dubbd/pull/322) to create the actual `uds-core-aws` package. This code is just a placeholder for that

.github/workflows/ci-on-pull-request-k3d.yaml
```diff
@@ -11,8 +11,8 @@ on:
       - "**.json"
       - "aws/**"
       - ".github/workflows/ci-on-pull-request-aws.y*ml"
-      - ".github/workflows/test-aws-package.y*ml"
-      - ".github/workflows/test-aws-upgrade.y*ml"
+      - ".github/workflows/**-aws-**.y*ml"
+      - ".github/workflows/**-aws-**.y*ml"
```
lucasrod16, line 15: This looks like a duplicate of the line right above it. What do you think about removing the duplicate and updating this pattern to use the following?:

```suggestion
      - ".github/workflows/**aws**.y*ml"
```

The current `**-aws-**` pattern won't handle `.github/workflows/ci-on-pull-request-dubbd-aws.yaml` and `.github/workflows/ci-on-pull-request-uds-core-aws.yaml` because they do not have a trailing `-` after `aws`

UncleGedd: thank you!

mjnagel (commented):
nit: `aws/uds-core-aws`/`aws/dubbd-aws` feels redundant, could we just have the folders be `aws/uds-core` and `aws/dubbd`?

aws/uds-core-aws/zarf.yaml
mjnagel: Works for me - noting this is a temp addition to get the structure.

UncleGedd: > (and ideal future state this might be importing from the skeleton, or something like that?)

Absolutely! Currently, the skeletons and the way we do TF in the `zarf.yaml` is [incompatible](https://github.com/defenseunicorns/zarf/issues/1808), so I think we have to stick with the monorepo pattern for now

.github/workflows/ci-on-pull-request-k3d.yaml
```diff
@@ -11,8 +11,8 @@ on:
       - "**.json"
       - "aws/**"
       - ".github/workflows/ci-on-pull-request-aws.y*ml"
-      - ".github/workflows/test-aws-package.y*ml"
-      - ".github/workflows/test-aws-upgrade.y*ml"
+      - ".github/workflows/**-aws-**.y*ml"
```
lucasrod16, line 14: This would be a duplicate now since `**aws**` also covers this case. We should be good to remove it

```suggestion
```

aws/dubbd-aws/zarf-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+      # -- Name of existing Terraform state bucket
+      state_bucket_name: uds-dev-state-bucket
+      # -- Key path to Terraform state file within the bucket
+      state_key: tfstate/dev/uds-dev-state-bucket.tfstate
+      # -- Name of DynamoDB table used for Terraform state locking
+      state_dynamodb_table_name: uds-dev-state-dynamodb
+      # -- AWS region
+      region: us-west-2
+      # -- Provision a private/internal load balancer for the tenant ingress gateway, if false a public load balancer will be provisioned
+      private_tenant_lb: true
```
MxNxPx, line 30: is this config intentionally missing this var?

```yaml
      # -- Provision a private/internal load balancer for the admin ingress gateway, if false a public load balancer will be provisioned
      private_admin_lb: true
```

.github/workflows/ci-on-pull-request-uds-core-aws.yaml
```diff
@@ -0,0 +1,26 @@
...
+  workflow_call:
+
+jobs:
+    test-release:
+        uses: ./.github/workflows/test-uds-core-aws-install.yaml
+        secrets: inherit
+
+    test-upgrade:
+        uses: ./.github/workflows/test-uds-core-aws-upgrade.yaml
+        secrets: inherit
```
mjnagel, line 26: This will fail until uds-core is published. Should we just comment these lines out?

aws/dubbd-aws/zarf-config.yaml
UncleGedd: Nope, my mistake, adding it back now

.github/workflows/ci-on-pull-request-uds-core-aws.yaml
UncleGedd: yep, good idea!

.github/workflows/ci-on-pull-request-k3d.yaml
UncleGedd: resolved and thanks for the catch

.github/workflows/ci-on-pull-request-uds-core-aws.yaml
mjnagel: :doh: just saw your other comment, happy to just accept it as failed for now.

UncleGedd:
> `aws/uds-core-aws`/`aws/dubbd-aws` feels redundant, could we just have the folders be `aws/uds-core` and `aws/dubbd`?

Happy to refactor specific folder names as we all deem fit, but I'd like to make this a follow-up discussion. I think we have a need to get this refactor merged in so we can continue with the `uds-core-aws` package development

lucasrod16 (approved):
lgtm!

PR #339, opened 2023-06-26
feat: add cloudtrail

MxNxPx:
Part of https://github.com/defenseunicorns/uds/issues/87 - does not integrate into UDS Core.


aws/cloudtrail/zarf-config.yaml
```diff
@@ -0,0 +1,20 @@
+package:
```
justin-o12, line 1: From my experience, these configs do not take affect but Zarf instead honors the root `zarf-config.yaml` that, in this case, would be found in `uds-core-aws/zarf-config.yaml`.

aws/cloudtrail/zarf.yaml
```diff
@@ -0,0 +1,98 @@
...
+  description: "Name of the pre-existing Terraform state S3 bucket"
+- name: STATE_KEY
+  description: "Path to the state file key in the state bucket"
+- name: STATE_DYNAMODB_TABLE_NAME
+  description: "Name of the DynamoDB table used for Terraform state locking"
+- name: REGION
+  description: "The AWS region to run the Terraform in"
+
+components:
+  - name: download-terraform
```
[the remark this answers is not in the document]
mjnagel: Yeah shouldn't be needed - we had added it to make a "standalone" cloudtrail package but technically terraform isn't even being used here. Okay with removing if that's the recommendation.

aws/cloudtrail/zarf-config.yaml
mjnagel: This was also meant to just be config for the "standalone" package, but if we don't want that, again okay with removing.

aws/cloudtrail/terraform.tfvars
```diff
@@ -0,0 +1 @@
+name          = "###ZARF_VAR_CLOUDTRAIL_NAME###"
```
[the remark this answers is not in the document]
mjnagel: @MxNxPx can chime in, but I think that was intentional - we made it pretty opinionated for multiregion, auto-created bucket, and 365 day retention. If we want to make these end user configurable we can add though for sure.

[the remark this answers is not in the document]
justin-o12: Oh, I see now.  You are saying to not use an external S3 bucket and therefore wouldn't need a bucket name.  Sounds good to me and thanks for the clarification.

aws/cloudtrail/zarf-config.yaml
MxNxPx: @justin-o12 -- i agree that primarily we would expect to use the config at the uds-core-aws level.  although i feel the benefit of splitting out some of these components is so they can be tested separately.  assuming we want to use zarf as part of the testing, then it probably is good to have a `zarf-config.yaml` for each (ex: eks, cloudtrail, etc) so they could be tested in isolation from others, as well as the uds-core-aws one.

aws/cloudtrail/zarf.yaml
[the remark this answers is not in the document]
MxNxPx: imho - the whole download-terraform ought to be looked at, but that is probably something for a different PR i would suggest.  this would be needed if we want to support the isolation of some of these components so they can be tested independently.  so i would suggest leaving it as it is, but open to other thoughts.

aws/cloudtrail/zarf-config.yaml
justin-o12: All good points and don't really disagree.  The other consideration is we could test these with [terratest like we're doing in this commit](https://github.com/defenseunicorns/uds-package-dubbd/pull/322/commits/df585ec68c54a368067e9f25176a7e8cc3fcae6d#diff-9057ddddd537bcb7e87bd4e8f04dcec4e61e31b21e74a32472865fd845de76b0).  I'm not sure the best way to go, either.  Maybe there is some advantage to testing this with Zarf to flex the variables.

mjnagel: We didn't add a CI workflow for this yet - happy to use terratest and borrow from the work you guys did for the workflow instead if we want to norm on that for testing these independent pieces.

It's an interesting pattern for zarf - `zarf.yaml`s that are just used for imports and don't really stand on their own as a zarf package.

aws/cloudtrail/main.tf
```diff
@@ -0,0 +1,96 @@
...
+    }
+    actions = [
+      "kms:Encrypt*",
+      "kms:Decrypt*",
+      "kms:ReEncrypt*",
+      "kms:GenerateDataKey*",
+      "kms:Describe*",
+    ]
+
+    resources = ["*"]
```
UncleGedd, line 48: I'm not really concerned with this given that we're pausing UDS Core AWS work, but I think we'd want specify to resources here instead of the wildcard

mjnagel: I'll take a look at this and see if its a quick fix - we had copied most of this from the example in the module.

mjnagel: Resolving per discussion - this can be tackled in a follow on issue.

mjnagel: Note that https://github.com/defenseunicorns/terraform-aws-uds-kms/issues/5 is a follow on issue that should influence this.

UncleGedd (approved):

PR #343, opened 2023-06-27
feat: Add Velero S3 bucket

cmharden:
Refactored loki terraform to be more generic. Added s3 bucket for Velero to dubbd-aws

Assumptions:
Loki and Velero should have independent life cycles where there corresponding data can be managed independently.


wirewc:
The PR looks decent overall. :+1: pending CI checks.

aws/dubbd-aws/zarf-config.yaml
```diff
@@ -18,19 +18,31 @@ package:
...
-      state_key: tfstate/dev/uds-dev-state-bucket.tfstate
       # -- Name of DynamoDB table used for Terraform state locking
       state_dynamodb_table_name: uds-dev-state-dynamodb
-      # -- AWS region
-      region: us-west-2
+      
+      # -- Key path to Terraform state file within the bucket
+      state_key_loki: tfstate/dev/uds-dev-state-bucket-loki.tfstate
+      # -- Key path to Terraform state file within the bucket
+      state_key_velero: tfstate/dev/uds-dev-state-bucket-velero.tfstate
```
UncleGedd, line 32: If possible, I don't think we should create a new config for this. Couple thoughts:
1. handle the key naming internally
2. use the same key for both Loki and Velero

aws/dubbd-aws/zarf.yaml
```diff
@@ -19,14 +19,16 @@ variables:
...
-- name: STATE_KEY
-  description: "Path to the state file key in the state bucket"
 - name: STATE_DYNAMODB_TABLE_NAME
   description: "Name of the DynamoDB table used for Terraform state locking"
-- name: REGION
-  description: "The AWS region to run the Terraform in"
+- name: STATE_KEY_LOKI
+  description: "Path to the state file key in the state bucket"
+- name: STATE_KEY_VELERO
+  description: "Path to the state file key in the state bucket"
```
UncleGedd, line 31: same comment as above regarding exposing another config

aws/dubbd-aws/zarf-config.yaml
```diff
@@ -18,19 +18,31 @@ package:
...
       #loki_kms_key_arn: "arn:aws:kms:us-west-2:000000000000:key/mrk-0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a"
       # -- If set to true, delete the S3 bucket and corresponding KMS key associated with the Loki bucket
-      #loki_force_destroy: "true"
\ No newline at end of file
+#     loki_force_destroy: "true"
+      
+      # -- Bring your own kms key, if omitted a key will be created with an alias prefix of "<cluster name>-velero-"
+      #velero_kms_key_arn: "arn:aws:kms:us-west-2:000000000000:key/mrk-0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a"
+      # -- If set to true, delete the S3 bucket and corresponding KMS key associated with the Velero bucket
+#     velero_force_destroy: "true"
```
UncleGedd, line 47: I think at this point we have a single config, called it `force_destroy` or maybe `ephemeral` that enables force destroying on both buckets in this case

aws/dubbd-aws/zarf.yaml
```diff
@@ -135,6 +143,75 @@ components:
...
+      target: run/velero/.terraform
+    - source: ../s3/main.tf
+      target: run/velero/main.tf
+    - source: ../s3/variables.tf
+      target: run/velero/variables.tf
+    - source: ../s3/output.tf
+      target: run/velero/output.tf
+    - source: ../s3/tf-velero.tfvars
+      target: run/velero/tf-velero.tfvars
+  - name: velero-terraform-execute
```
UncleGedd, line 164: To reduce the amount of new code here, I think we could combine the loki and velero components that create S3 buckets into a single component. Thoughts?

UncleGedd:
Generally speaking, I love the idea of refactoring that `loki` folder into a generic `s3` folder. I think we might be able to apply that same sort of thinking to reduce the amount of new code and configs for adding a Velero bucket

aws/dubbd-aws/zarf-config.yaml
cmharden: Updated to handle naming internally, but solution assumes sed will be installed.

aws/dubbd-aws/zarf-config.yaml
cmharden: Updated to use a single `ephemeral` variable to enable `force_destroy` for both loki and velero.

Issue #337, opened 2023-06-26
Decouple S3 and IRSA


UncleGedd:

wirewc:
@UncleGedd can you add the UDS team as an admin to the [irsa](https://github.com/defenseunicorns/terraform-aws-uds-irsa) repository please?

wirewc:
UDS is now listed as maintainer. Probably good enough unless anything with the pipeline needs to be modified or renovate needs to be added.

wirewc:
@lucasrod16 has a jump start on this. There was a little collaboration to help understand and get the basics in place for the terraform repository structure. 

**Not for future work:** Document basic IaC structure for new repositories.

wirewc:
After speaking with @zachariahmiller, he mentioned that a prior engagement used this module. https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/modules/irsa. This module is heavily used and recommended for people working in secure environments. I believe we have used EKS blueprints in the EKS module already.

Question for @UncleGedd and @lucasrod16. The IRSA module seems simple enough and this is maintained, however I am not seeing any tests shared or reports of any scans. Alternatively, we could call up the upstream module and perform scans as needed to get the tfsec and regula information that should satisfy the needs of an auditor.

@CloudBeard from a auditor perspective, do you have a comment on AWS Blueprints versus us rolling our own (and testing it)?

noahpb:
version `0.0.5` of the s3 [module](https://github.com/defenseunicorns/terraform-aws-uds-s3/releases/tag/v0.0.5) removes irsa resources

lucasrod16:
Fixed by https://github.com/defenseunicorns/terraform-aws-uds-s3/pull/20

PR #384, opened 2023-07-10
fix: Update loki terraform to use generic IRSA module

lucasrod16:
Closes https://github.com/defenseunicorns/uds-package-dubbd/issues/337


mjnagel:
I think we may be able to close this out since similar changes were added in #420

noahpb:
> 

Just now seeing this, you are correct @mjnagel. I will close.

Issue #385, opened 2023-07-11
Modify `loki_force_destroy` to be tied to generic `ephemeral` var


mjnagel:
As we add additional persistence/features to DUBBD we will want to allow for toggling whether the user wants the cluster/storage to be `ephemeral` or not. Having this tied to a single var will make this easier for users.

PR #388, opened 2023-07-11
Tie force destroy to ephemeral var

mjnagel:
Intentional decision to not just rename the variable:
- Allows for non-"true/false" things tied to ephemeral (we can "translate" the true/false ephemeral to any necessary "type")
- Allows for controlling the individual pieces on there own still if desired (maybe user wants to delete PVCs but keep Loki S3)

Closes https://github.com/defenseunicorns/uds-package-dubbd/issues/385


MxNxPx (approved):
:robot: likey!

PR #403, opened 2023-07-15
fixes k3d local installation

UncleGedd:
- Fix: metal-lb's speaker pod was failing due to not being able to access ConfigMaps
- Fix: Promtail pod wouldn't start because it couldn't access `/etc/machine-id`
- Removes extraneous args from the `k3d` command
- Adds artifacts to k3d .gitignore

### todo
- make sure we are dogfooding `k3d/local`?

### DIFF
--- a/k3d/local/manifests/metallb/metallb-native.yaml
+++ b/k3d/local/manifests/metallb/metallb-native.yaml
@@ -1,3 +1,4 @@
+# sourced from: https://raw.githubusercontent.com/metallb/metallb/v0.13.10/config/manifests/metallb-native.yaml
 apiVersion: v1
 kind: Namespace
 metadata:
@@ -12,6 +13,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.11.1
+  creationTimestamp: null
   name: addresspools.metallb.io
 spec:
   conversion:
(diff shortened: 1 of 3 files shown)


k3d/local/manifests/metallb/metallb-native.yaml
```diff
@@ -1308,6 +1308,7 @@ rules:
   - ""
   resources:
   - secrets
+  - configmaps
```
mjnagel, line 1311: I'm curious on the error you saw here. This file comes straight from the upstream source I believe (https://raw.githubusercontent.com/metallb/metallb/v0.13.10/config/manifests/metallb-native.yaml) and also does not have configmaps upstream.

k3d/local/scripts/k3d.sh
```diff
@@ -12,8 +12,9 @@ else
     --k3s-arg "--disable=traefik@server:*" \
     --k3s-arg "--disable=metrics-server@server:*" \
     --k3s-arg "--disable=servicelb@server:*" \
-    --port 443:443@loadbalancer \
-    --port 80:8080@loadbalancer \
+    --volume /tmp/dubbd:/var/lib/rancher/k3s/storage@all \
+    --volume /etc/machine-id:/etc/machine-id \
+    --volume /var/log/pods:/var/log/pods \
```
mjnagel, line 17: Also curious on this one...I've deployed with k3d without mounting this extra volume and saw no issues with promtail. It does normally log a warning that certain paths are directories vs log files but should still tail the log files okay?

UncleGedd: You're right, this was leftover from testing, removed with latest commit

mjnagel (approved):
LGTM! Thanks for finding and fixing these issues.

Issue #379, opened 2023-07-07
Kyverno CPU resource limits too low


brianrexrode:
While doing some recent testing on DUBBD deployments, we ran into issues with kyverno pod resource limits being exceeded and causing pod restarts which in turn is causing other BB pods to be stuck in init status with the following errors when doing a describe on the pods:
```
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "5ba287e76728b49f7ce99940d01d6b8e274b35c14a7c0daa2a531c3a067383e8": plugin type="aws-cni" name="aws-cni" failed (add): add cmd: failed to assign an IP address to container
```
In some cases these errors clear themselves and some pods finally initialize and become ready. In other cases the pods remain in init status ultimately causing the associated helmrelease to fail which eventually causes the zarf pkg deploy to fail. Also found numerous errors like this having issues communicating with kyverno svc:
```
Internal error occurred: failed calling webhook \"validate.kyverno.svc-fail\": failed to call webhook: Post \"[https://kyverno-kyverno-svc.kyverno.svc:443/validate/fail?timeout=30s](https://kyverno-kyverno-svc.kyverno.svc/validate/fail?timeout=30s)\": context deadline exceeded"}
```
I was able to manually increase the cpu/memory resource limits for kyverno (I doubled them) and tested the deployment again to ensure all pods came up. While running kubectl top pods -n kyverno we noticed that the pods CPU usage was upwards of 1000m (BB charts default to 500m). 
Found this [issue](https://github.com/kyverno/kyverno/issues/5485#issuecomment-1422468133) which identified a memory leak in kyverno v1.8.5.

brianrexrode:
It looks like BB 2.6.0 will include the kyverno upgrade to v1.9.2 hopefully fixing the memory leak issue

oates:
@MxNxPx you got this one or tracking?

brianrexrode:
I was able to test a deployment of DUBBD that included BB 2.6.0 and can confirm the deployment did complete successfully. kyverno pods did experience some CPU spikes but decreased CPU usage quickly. 

I would consider this issue `Closed` as soon as a new release is cut that includes BB 2.6.0.

Thanks

MxNxPx:
closing as [this dubbd release](https://github.com/defenseunicorns/uds-package-dubbd/pull/400) includes BB 2.6.0

PR #411, opened 2023-07-17
feat: add dubbd pre-upgrade & update refs

MxNxPx:
prereq for https://github.com/defenseunicorns/uds-package-dubbd/pull/401 & https://github.com/defenseunicorns/uds-package-dubbd/issues/379

### DIFF
--- a/defense-unicorns-distro/zarf.yaml
+++ b/defense-unicorns-distro/zarf.yaml
@@ -79,21 +79,6 @@ components:
               sensitive: true
         - cmd: rm cat_certs.sh
         - cmd: rm cat_key.sh
-  - name: preflight
-    required: true
-    actions:
-      onDeploy:
-        before:
-        - cmd: "./zarf tools kubectl version --short || true"
-          setVariables:
-            - name: OUTPUT
-        after:
-        - cmd: ./preflight.sh
-        - cmd: rm preflight.sh
-    files:
-    - source: scripts/preflight.sh
-      target: preflight.sh
-      executable: true
   - name: download-flux
     required: true
     actions:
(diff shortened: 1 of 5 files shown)


defense-unicorns-distro/zarf.yaml
```diff
@@ -121,6 +106,41 @@ components:
...
+      onDeploy:
+        before:
+        - cmd: |
+            if [ "${ZARF_VAR_IS_UPGRADE}" = "true" ]; then
+              echo "Performing pre-upgrade steps" >/dev/null 2>&1
+            fi
+          description: "Perform DUBBD pre-upgrade steps"
+        - cmd: |
+            if [ "${ZARF_VAR_IS_UPGRADE}" = "true" ]; then
+              ./zarf tools kubectl delete helmrelease -n bigbang kyverno-policies --ignore-not-found --timeout=20m
```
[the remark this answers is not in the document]
mjnagel: Going to resolve this convo - per testing with @MxNxPx this didn't resolve the issue unfortunately 😢

defense-unicorns-distro/zarf.yaml
```diff
@@ -121,6 +106,41 @@ components:
...
+          description: "Detect if this is an upgrade"
+          setVariables:
+            - name: IS_UPGRADE
+        after:
+        - cmd: ./preflight.sh
+        - cmd: rm preflight.sh
+    files:
+    - source: scripts/preflight.sh
+      target: preflight.sh
+      executable: true
```
mjnagel, line 128: I came up with this great idea all on my own, definitely not an idea from @MxNxPx - while we're moving around components it might be good to make preflight the very first component so that users get instant feedback if they are attempting to deploy on an unsupported cluster version.

defense-unicorns-distro/zarf.yaml
```diff
@@ -121,6 +106,41 @@ components:
...
+  - name: dubbd-pre-upgrade
+    required: true
+    actions:
+      onDeploy:
+        before:
+        - cmd: |
+            if [ "${ZARF_VAR_IS_UPGRADE}" = "true" ]; then
+              echo "Performing pre-upgrade steps" >/dev/null 2>&1
+            fi
+          description: "Perform DUBBD pre-upgrade steps"
```
mjnagel, line 138: I'd recommend we just drop this portion in favor of just using the component name to identify this as the pre-upgrade steps. Per testing an empty component will function fine and still allow the flexibility to extend/add actions as needed.

defense-unicorns-distro/zarf.yaml
MxNxPx: this sounds like great advice you came up with all on your own @mjnagel, implementing now.

defense-unicorns-distro/zarf.yaml
MxNxPx: agree @mjnagel - dropping this should reduce confusing output

defense-unicorns-distro/zarf.yaml
```diff
@@ -214,3 +234,13 @@ components:
...
+  - name: dubbd-post-upgrade
+    required: true
+    actions:
+      onDeploy:
+        before:
+        - cmd: |
+            if [ "${ZARF_VAR_IS_UPGRADE}" = "true" ]; then
+              echo "Performing post-upgrade steps" >/dev/null 2>&1
+            fi
+          description: "Perform DUBBD post-upgrade steps"
```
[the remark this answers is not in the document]
MxNxPx: good catch, removing this as well.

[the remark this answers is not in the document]
MxNxPx: good catch, removing this as well.

mjnagel (approved):
🥳

Issue #348, opened 2023-06-29
Configure default storage class


UncleGedd:
As a DUBBD user
If a PersistentVolume gets deleted
I don't want the underlying EBS volume to get deleted
So that I am less likely to lose data

mjnagel:
Is this targeted for "DUBBD-AWS" or "UDS Core-AWS"? I lean towards the latter - if you are bringing your own cluster (DUBBD-AWS) then you likely would bring your own default storage class too?

I assume we want this as a zarf var, maybe tied to existing loki deletion as well (switch to generic `ephemeral` var)?

UncleGedd:
Good question, I was actually thinking this would target both packages, but there may be more to this ticket than I originally thought. Do you think we should discuss what type of StorageClass we could be opinionated on?

mjnagel:
Yeah I think it's worth discussing. My initial 2cents would be that we setup something like this:
```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: gp2
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
allowVolumeExpansion: true
provisioner: kubernetes.io/aws-ebs
volumeBindingMode: WaitForFirstConsumer
parameters:
  type: gp2
reclaimPolicy: ###ZARF_VAR_SOMETHING###
```

Opinions we would have here:
- our storage class should be the default - self explanatory and makes sure that we don't have to deal with configuring all the DUBBD things to use it.
- volume expansion should be allowed - I think this is best especially while we are relying on PVCs for grafana, prometheus, etc, which the end user might want to resize eventually.
- volumes shouldn't be bound until used - feels better to be slower than to over-provision volumes that might not be used?
- ebs over efs - efs is far more expensive and I don't think we have a use case for this yet - might be worth exploring as an alternative (non-default) storage class in the future if the need arises.
- gp2 over gp3 - I don't have a strong opinion on this one...gp3 can be more cost effective and/or more performant in some scenarios but I lean towards gp2 as a more generic baseline?

User configurable options:
- reclaim policy - really this would just be for dev vs prod ideally, along the same lines as exposing the loki force delete of S3. I think the idea of a general `ephemeral` value has been floated around, which might control all of these knobs together.

mjnagel:
[...]

PR #417, opened 2023-07-18
feat: add default storage class for DUBBD

mjnagel:
Notable things:
- gp3 storage class for all DUBBD things *if* the EBS CSI driver is installed
- if on UDS Core, this is also the cluster default
- upgrade process leaves things as is in order to be non-disruptive for now (this could change with automation around velero, but would be a larger effort)
- zarf var is used to determine the reclaim policy (based off ephemeral value)
- if the storage class is being used and set to ephemeral PVCs will be deleted on removal

Closes https://github.com/defenseunicorns/uds-package-dubbd/issues/348


mjnagel:
Moving to draft temporarily to add docs/ADR to this.

MxNxPx (approved):
:shipit:

PR #420, opened 2023-07-19
Update S3/irsa iac for Loki/Velero

cmharden:


aws/dubbd-aws/zarf.yaml
```diff
@@ -40,6 +40,8 @@ variables:
   default: "false"
 - name: LOKI_FORCE_DESTROY
   description: "Provide a means to force destroy the Loki storage (S3) bucket even if it contains data, overrides ephemeral setting"
+- name: VELERO_FORCE_DESTROY
+  description: "Provide a means to force destroy the Loki storage (S3) bucket even if it contains data, overrides ephemeral setting"
```
mjnagel, line 44:
```suggestion
  description: "Provide a means to force destroy the Velero storage (S3) bucket even if it contains data, overrides ephemeral setting"
```

aws/dubbd-aws/zarf-config.yaml
```diff
@@ -36,3 +36,8 @@ package:
       #loki_kms_key_arn: "arn:aws:kms:us-west-2:000000000000:key/mrk-0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a"
       # -- If set to true, delete the S3 bucket and corresponding KMS key associated with the Loki bucket. Overrides ephemeral setting.
       #loki_force_destroy: "true"
+
+      # -- Bring your own kms key, if omitted a key will be created with an alias prefix of "<cluster name>-velero-"
+      #velero_kms_key_arn: "arn:aws:kms:us-west-2:000000000000:key/mrk-0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a"
+      # -- If set to true, delete the S3 bucket and corresponding KMS key associated with the Velero bucket. Overrides ephemeral setting.
+      #Velero_force_destroy: "true"
```
mjnagel, line 43: @MxNxPx I think we wanted to get away from having the individual vars for destruction exposed in the config file, and just have the `ephemeral` setting front and center, is that right?

I was making part of that change in https://github.com/defenseunicorns/uds-package-dubbd/pull/417/files but removing this one can always be a cleanup in follow on PR if everything else is good to go.

aws/dubbd-aws/zarf.yaml
```diff
@@ -60,6 +62,14 @@ components:
             mute: true
             setVariables:
               - name: LOKI_FORCE_DESTROY
+          - cmd: |
+              if [ -z "${ZARF_VAR_VELERO_FORCE_DESTROY}" ]; then
+                echo "${ZARF_VAR_EPHEMERAL}"
+              else
+                echo "${ZARF_VAR_VELERO_FORCE_DESTROY}"
+              fi
+            setVariables:
```
mjnagel, line 71: nit: preferable to set mute here for consistency and cleaner output.

```suggestion
            mute: true
            setVariables:
```

aws/dubbd-aws/zarf.yaml
```diff
@@ -110,17 +122,19 @@ components:
...
       target: run/loki/terraform.tfvars
-  - name: execute-terraform
+  - name: loki-execute-terraform
     required: true
     actions:
       onDeploy:
         before:
-          - cmd: "test -d ./loki/ && chmod -R ugo+rwx ./loki/ || echo $?"
-            dir: run
+          - cmd: "echo ${ZARF_VAR_STATE_KEY} | sed 's/\.tfstate/-loki.tfstate/g'"
```
mjnagel, line 130: By changing construction of the state key are we breaking upgrades? Or will things migrate to the new state seamlessly?

(we might not even care, just asking to make sure I understand what is happening here)

aws/dubbd-aws/zarf.yaml
```diff
@@ -155,10 +169,80 @@ components:
...
+              -backend-config="bucket=${ZARF_VAR_STATE_BUCKET_NAME}" \
+              -backend-config="key=${ZARF_VAR_STATE_KEY_VELERO}" \
+              -backend-config="region=${ZARF_VAR_REGION}" \
+              -backend-config="dynamodb_table=${ZARF_VAR_STATE_DYNAMODB_TABLE_NAME}"
+            dir: run/velero
+          - cmd: ../terraform plan --out=velero.out
+            dir: run/velero
+          - cmd: sleep 15 #time to review
+          - cmd: ../terraform apply velero.out 
+            dir: run/velero
```
mjnagel, line 215: I don't know which is *best* but I would recommend we stay consistent across the way each of loki/velero is being applied (loki just plans with no output, then auto-approve applies vs the output here).

aws/values/aws-velero.yaml
```diff
@@ -0,0 +1,119 @@
...
+                cpu: 100m
+              limits:
+                memory: 512Mi
+                cpu: 100m
+      enableHelmHooks: true
+      extraVolumeMounts: [ ]
+      extraVolumes: [ ]
+      image:
+        repository: registry1.dso.mil/ironbank/opensource/velero/velero
+        tag: v1.10.2
```
mjnagel, line 66: A lot of the config in here looks like it matches the BB defaults (especially the image references). Where possible we should try to inherit those types of things and not "override" so that we have less maintenance on our side when a BB update (containing a velero update) gets published.

Also we can probably clean this up a lot by removing any of the `key: [ ]`, `key: { }`, and `key: null` values. That will help with readability and being able to better understand what it is that we are actually configuring.

aws/dubbd-aws/zarf.yaml
cmharden: Fixed

aws/dubbd-aws/zarf-config.yaml
cmharden: Removed

aws/dubbd-aws/zarf.yaml
cmharden: Fixed

aws/dubbd-aws/zarf.yaml
cmharden: :+1:

aws/values/aws-velero.yaml
cmharden: Cleaned up

aws/dubbd-aws/zarf.yaml
cmharden: Live convo - issue resolved

mjnagel: Just for others reading this later - determined the upgrade considerations here are not worth resolving, this will be a "breaking change" but keeps our keys together under a single dir which is more desired.

aws/values/aws-velero.yaml
```diff
@@ -0,0 +1,31 @@
...
+        volumeSnapshotLocation:
+          config:
+            region: "###ZARF_VAR_VELERO_S3_AWS_REGION###"
+          provider: aws
+      plugins:
+        aws:
+          image:
+            repository: registry1.dso.mil/ironbank/opensource/velero/velero-plugin-for-aws
+      image:
+        repository: registry1.dso.mil/ironbank/opensource/velero/velero
```
mjnagel, line 23: Should be safe to remove these lines as well, they are defaults in the chart I believe.

aws/values/aws-velero.yaml
```diff
@@ -0,0 +1,31 @@
...
+      serviceAccount:
+        server:
+          annotations:
+            eks.amazonaws.com/role-arn: "###ZARF_VAR_VELERO_S3_ROLE_ARN###"
+      kubectl:
+        image:
+          repository: registry1.dso.mil/ironbank/opensource/kubernetes/kubectl
```
mjnagel, line 30: Should be safe to remove these lines as well, they are defaults in the chart I believe.

aws/values/aws-velero.yaml
mjnagel:
```suggestion
```

aws/values/aws-velero.yaml
mjnagel:
```suggestion
```

aws/loki/main.tf
```diff
@@ -23,48 +23,109 @@ data "aws_region" "current" {}
...
+  kms_key_description = "${var.name} DUBBD deployment Loki Key" # Description for the KMS key.
+  tags = {
+    Deployment = "UDS DUBBD ${local.name}"
   }
 }
+
+
+module "irsa" {
+  source                        = "github.com/defenseunicorns/terraform-aws-uds-irsa?ref=v0.0.2"
+  name                       = "${local.name}-irsa"
```
noahpb, line 88: nit: This is on me, but the "-irsa" suffix is redundant as the irsa module automatically [injects](https://github.com/defenseunicorns/terraform-aws-uds-irsa/blob/db33cbc96b26d5f9265c5e026593bcb598b84c08/main.tf#L2) "irsa" into the role name. The value for name here is [expected](https://github.com/defenseunicorns/terraform-aws-uds-irsa/blob/db33cbc96b26d5f9265c5e026593bcb598b84c08/variables.tf#L9) to be the cluster name. `local.name` would suffice

MxNxPx (approved):

mjnagel (approved):
:shipit:

Issue #304, opened 2023-06-20
Secure default authentication/credentials


mjnagel:
Currently DUBBD does no configuration for default credentials/authentication. Grafana and NeuVector are configured with default admin user/pass configs that would be insecure in a production environment. 

We have a number of different options for how to handle this:
- Require user to pass in zarf vars to configure user/pass for default admin
- Randomly set a user/pass for default admin (and add to secret somewhere?)
- Disable default admin (is this possible?)
- Document a post-install step for user to setup/change default credentials

noahpb:
Added auto-generation of default creds for grafana in https://github.com/defenseunicorns/uds-package-dubbd/commit/b4ec6a71eda00a5e373a8a34d13e916d3954e549

PR #472, opened 2023-08-04
feat: non default creds for admin interfaces

noahpb:
Addresses Grafana and NeuVector UI for #304 

Behavior:
- Changes Grafana and NeuVector admin password from the BB default to a randomly generated string
- Creates an `admin-interface-credentials` secret in the `zarf` namespace where all admin UI credentials can be queried
- Does not change password on upgrade (plan to support that in a later PR via optional parameter)
- Credential for Grafana is passed as a string value to grafana helm chart parameter `adminPassword`
- Credentials for NeuVector are passed into NeuVector helm chart parameter `userinitcfg.yaml` (overrides [values](https://repo1.dso.mil/big-bang/bigbang/-/blob/master/chart/templates/neuvector/values.yaml#L33) already declared in BB, thus forcing this PR to address creating those values as well)


defense-unicorns-distro/manifests/admin-interface-credentials.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: dubbd-admin-interface-credentials
+  namespace: zarf
+  labels:
+    generated-by: dubbd
+data:
+  grafana-password: ###ZARF_VAR_GRAFANA_ADMIN_PASSWORD_BASE64###
```
mjnagel, line 9: Is there a reason to use `data` over `stringData`? Just thinking simplification of vars, this could just be a single zarf var for the plaintext password if we switched to `stringData` I think (and then we wouldn't need the additional action to base64 encode it either).

noahpb: Great suggestion, I think that is much cleaner. Addressed in [541be5d](https://github.com/defenseunicorns/uds-package-dubbd/pull/472/commits/541be5d756ffe305892049c5382f791c2f8a4741).

mjnagel (commented):
Overall looks great - might be nice to add a doc addition _somewhere_ with a command to grab the grafana password?

noahpb:
> Overall looks great - might be nice to add a doc addition _somewhere_ with a command to grab the grafana password?

Good point. Addressed in https://github.com/defenseunicorns/uds-package-dubbd/pull/472/commits/9398ac76952887861df2fcea2059019daaa15e6b

MxNxPx (dismissed):
:shipit:

noahpb:
as discussed, moving this to a draft state as work for the other components mentioned in #304 is continued. FYI @mjnagel @MxNxPx

defense-unicorns-distro/zarf.yaml
```diff
@@ -120,6 +120,40 @@ components:
...
+            mute: true
+            setVariables:
+              - name: GRAFANA_ADMIN_PASSWORD
+                sensitive: true
+                autoIndent: true
+          - cmd: "openssl rand -base64 24"
+            mute: true
+            setVariables:
+              - name: NEUVECTOR_METRICS_PASSWORD
+                sensitive: true
```
noahpb, line 138: @rjferguson21 @mjnagel in these lines, i am recreating the logic in BB ([ref](https://repo1.dso.mil/big-bang/bigbang/-/blob/master/chart/templates/neuvector/values.yaml#L37)) for randomly generating a password for the `metrics` user account in neuvector (due to that fact that I need to merge an additional value into the `userinitcfg.yaml` section). 

As designed currently, this value will be regenerated each time a dubbd package is upgraded. Would this have any effects on any downstream services? (ie, would a service need to be manually restarted to pick up the new credential after a dubbd upgrade?)

mjnagel: Yeeeep, so two pieces you'll need beyond what you have already:
- [passing the password into the exporter pod](https://repo1.dso.mil/big-bang/bigbang/-/blob/master/chart/templates/neuvector/values.yaml#L81) - this will pass the same password to the exporter, where it is used to login/metrics grabbing
- restart of exporter on password change - in BB we handled this with a [helm checksum, turned into a pod annotation](https://repo1.dso.mil/big-bang/bigbang/-/blob/master/chart/templates/neuvector/values.yaml#L60), but in the context of DUBBD it could just be an action on upgrade

mjnagel: (the controller will restart automatically for you based on a checksum, so you won't have to handle restarts anywhere else)

noahpb: Discussed offline, we are going to focus on not regenerating this value on each upgrade

noahpb: I believe everything in https://github.com/defenseunicorns/uds-package-dubbd/pull/472/commits/866fdd1569f39944bd1dc3ffa36903ccfc93cc76 and https://github.com/defenseunicorns/uds-package-dubbd/pull/472/commits/d85a60ed4bd420c07b202610d815e4e0de6ff263 addresses everything we talked about @mjnagel

defense-unicorns-distro/zarf.yaml
```diff
@@ -120,6 +120,37 @@ components:
                 autoIndent: true
           - cmd: rm bigbang.dev.cert
           - cmd: rm bigbang.dev.key
+  - name: generate-credentials
+    required: true
+    actions:
+      onDeploy:
+        before:
+          - cmd: "if [ ${ZARF_VAR_IS_UPGRADE} = true ]; then ./zarf tools kubectl get secret -n zarf dubbd-admin-interface-credentials -o go-template='{{index .data \"grafana-password\" | base64decode}}' ; else openssl rand -base64 24; fi"
```
mjnagel, line 128: Dropping comment here based on what we discussed synchronously - would request that these become multi-line cmd blocks (`cmd: |`) for better readability.

defense-unicorns-distro/zarf.yaml
```diff
@@ -120,6 +120,37 @@ components:
...
+            mute: true
+            setVariables:
+              - name: GRAFANA_ADMIN_PASSWORD
+                sensitive: true
+          - cmd: "if [ ${ZARF_VAR_IS_UPGRADE} = true ]; then ./zarf tools kubectl get secret -n neuvector neuvector-prometheus-exporter-pod-secret -o go-template='{{index .data \"CTRL_PASSWORD\" | base64decode}}' >/dev/null 2>&1; if [ $? -ne 0 ]; then openssl rand -base64 24; fi; else openssl rand -base64 24; fi"
```
mjnagel, line 133: As discussed synchronously I believe there's a bug here in the case of the secret already existing due to nulling stdout on the `kubectl get`.

defense-unicorns-distro/zarf.yaml
noahpb: addressed in https://github.com/defenseunicorns/uds-package-dubbd/pull/472/commits/90085ef3d365c702fc56cb7ee2872e48aaf72f36

defense-unicorns-distro/zarf.yaml
noahpb: addressed in https://github.com/defenseunicorns/uds-package-dubbd/pull/472/commits/90085ef3d365c702fc56cb7ee2872e48aaf72f36

mjnagel (approved):
:shipit: nice work on this @noahpb !!!

mjnagel (approved):

PR #481, opened 2023-08-07
fix: updating neuvector oscal

CloudBeard:
Updated NeuVector path which includes an update to the OSCAL.

*Note the NeuVector verison in use in this repo is ahead of upstream BB version becuase the current version does not include OSCAL. Should catch up in 2.8.0

### DIFF
--- a/.github/actions/create-zarf-package/action.yaml
+++ b/.github/actions/create-zarf-package/action.yaml
@@ -34,7 +34,7 @@ runs:
       uses: defenseunicorns/setup-zarf@main
       with:
         # renovate: datasource=github-tags depName=defenseunicorns/zarf versioning=semver
-        version: v0.28.3
+        version: v0.28.4
         download-init-package: ${{ inputs.download-init-package }}
 
     - name: Create Zarf Package
--- a/.github/workflows/ci-on-pull-request-dubbd-aws.yaml
+++ b/.github/workflows/ci-on-pull-request-dubbd-aws.yaml
@@ -15,6 +15,7 @@ on:
       - "aws/cloudtrail/**"
       - "aws/**/test/**"
       - "k3d/**"
+      - "rke2/**"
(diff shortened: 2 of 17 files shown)


mjnagel (commented):
Shouldn't this have updated the OSCAL as a result of the config change? 🤔 @CloudBeard

CloudBeard:
@mjnagel It should have, noticed after this the component generate dosen't check back matter. Ill have to update that first.

mjnagel:
@CloudBeard also wanted to note there's a new OSCAL component/config [here](https://github.com/defenseunicorns/uds-package-dubbd/blob/main/.github/workflows/oscal/config-dubbd-rke2-oscal-component.yaml) that will need the same change now - pending potentially moving the `rke2` folder into `defense-unicorns-distro`...

CloudBeard:
@mjnagel updated the RKE2 oscal for NeuVector change.

Still waiting to get that component generator updated to catch that diff in only the back matter.

MxNxPx:
@CloudBeard - i think this got addressed in the most recent release and this PR can be closed without merging.  Would you mind confirming?

CloudBeard:
@MxNxPx The main parts of the OSCAL are there so this can be closed. 

** Reason I left it open was I ran into a small issue that this PR discovered in the underlying component generator tool. If there is no change in the body the tool dosen't run a diff on just the back matter section. Waited to see level of effort on updating the tool and bumping this but needs to be a separate issue I track and fix since its going to take too long to get to.

mjnagel:
@MxNxPx if you see the diffs in https://github.com/defenseunicorns/uds-package-dubbd/pull/523 there is a change in the neuvector piece. That's the change we want everywhere, but it's currently only in the RKE2 package because of the issue @CloudBeard mentioned with the component-generator not diffing the backmatter section.

Issue #491, opened 2023-08-09
remove kiali operator from dubbd


MxNxPx:
As a delivery engineer I don't want to deploy the kiali operator that is used sparingly and consumes lots of cluster compute resources.

MxNxPx:
opened [ADR pr](https://github.com/defenseunicorns/uds-package-dubbd/issues/491)

MxNxPx:
will not remove kiali, instead need to fix kiali per https://github.com/defenseunicorns/uds-package-dubbd/issues/505

PR #493, opened 2023-08-10
docs: add adr for disabling kiali operator

MxNxPx:
addresses https://github.com/defenseunicorns/uds-package-dubbd/issues/491

### DIFF
--- a/docs/adrs/0003-disable-kiali-operator-for-dubbd.md
+++ b/docs/adrs/0003-disable-kiali-operator-for-dubbd.md
@@ -4,7 +4,7 @@ Date: 2023-08-09
 
 ## Status
 
-Accepted
+Rejected (keeping Kiali Operator per feedback received)
 
 ## Context
 
--- a/docs/adrs/0003-disable-kiali-operator-for-dubbd.md
+++ b/docs/adrs/0003-disable-kiali-operator-for-dubbd.md
@@ -1,0 +1,35 @@
+# 3. Disable Kiali Operator for DUBBD
+
+Date: 2023-08-09
+
+## Status
+
+Accepted
+
+## Context
+
+Big Bang core (umbrella) is configured with certain "defaults" such as charts enabled or disabled and specific application configurations. It is preferred that to stay aligned wherever possible with these defaults, but in certain scenarios DUBBD may deviate from them, for example:
+
+* Simplification
+* Performance optimization
+* Compliance / controls alignment
+* Best practices (from delivery experience)
+* Application features
+
+## Decision
+
+DUBBD will be configured to disable the Kiali Operator, which is currently enabled in Big Bang core by default.
+
+## Consequences
+
+This change is expected to improve DUBBD in these regards:
+
+* Simplification - less software to support and maintain
+* Performance optimization - lower cluster compute utilization as the Kiali Operator is resource intensive
+* Compliance / controls alignment - no impact to existing controls being satisfied by DUBBD
+* Best practices (from delivery experience) - anecdotally, the "always-on" Kiali instance was leveraged only occasionally, as opposed to other similar DUBBD tools that were used very regularly
+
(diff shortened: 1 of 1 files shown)


corang:
Is skipping the operator and deploying kiali directly an option here? I think it still provides value

MxNxPx:
> Is skipping the operator and deploying kiali directly an option here? I think it still provides value

@corang - i hope that i am wrong... but i think @mjnagel had said BB doesn't have straight Kiali available

mjnagel:
@MxNxPx @corang BB does not publish a Kiali only chart, but the image would be the same one from Ironbank. It would mean we have to maintain our own chart with the BB integration pieces (or some 🌶️ magic to do that). Not totally infeasible but not as easy as adding a new gitrepo in our values.

docs/adrs/0003-disable-kiali-operator-for-dubbd.md
```diff
@@ -0,0 +1,35 @@
...
+Accepted
+
+## Context
+
+Big Bang core (umbrella) is configured with certain "defaults" such as charts enabled or disabled and specific application configurations. It is preferred that to stay aligned wherever possible with these defaults, but in certain scenarios DUBBD may deviate from them, for example:
+
+* Simplification
+* Performance optimization
+* Compliance / controls alignment
+* Best practices (from delivery experience)
```
mjnagel, line 16: Recommend we remove "delivery" wording (here and a few other spots below) and switch to user/customer to use more public facing language here.

blancharda:
Some end users have expressed concern over (potentially) losing access to the tool.

They described it as a "single pane of glass" view for monitoring the current state of their deployments and in-cluster communication, which they use regularly during upgrade cycles to monitor health.

My preference would be to leave it in place until we have an alternative recommendation for installing it ad-hoc.
(And perhaps doing a broader user survey to better understand more end-user's usage patterns.)

That said, the underlying resource usage problem is still a problem..
+1 for @corang's suggestion above.. the [kiali-server chart](https://kiali.io/docs/installation/installation-guide/install-with-helm/) may be a potential starting point.

docs/adrs/0003-disable-kiali-operator-for-dubbd.md
MxNxPx: good call @mjnagel - replaced with the more general "platform" verbiage via [this commit](https://github.com/defenseunicorns/uds-package-dubbd/commit/50d6fcaf5e68c199de85eb1033101feda8f0825d)

corang (changes requested):
Leaving it in is fine, but we still haven't addressed the problem. Kiali is still of no value if it never gets deployed by the operator

corang:
> Leaving it in is fine, but we still haven't addressed the problem. Kiali is still of no value if it never gets deployed by the operator

It's probably the most work out of all the issues but it seems like deploying kiali ourself directly without the operator solves the most problems.

corang:
> > Leaving it in is fine, but we still haven't addressed the problem. Kiali is still of no value if it never gets deployed by the operator
> 
> It's probably the most work out of all the issues but it seems like deploying kiali ourself directly without the operator solves the most problems.

Or as a bandaid just bump the memory limit a few gigs for now and mark it as a known issue

MxNxPx:
> > > Leaving it in is fine, but we still haven't addressed the problem. Kiali is still of no value if it never gets deployed by the operator
> > 
> > 
> > It's probably the most work out of all the issues but it seems like deploying kiali ourself directly without the operator solves the most problems.
> 
> Or as a bandaid just bump the memory limit a few gigs for now and mark it as a known issue

@corang - agreed.  i have opened [this ticket](https://github.com/defenseunicorns/uds-package-dubbd/issues/505) to address separately from this ADR.

mjnagel (approved):

corang (approved):
Given the other ticket this is fine

PR #499, opened 2023-08-14
feat: increase kyverno requests/limits

noahpb:
bumps requests and limits for kyverno


mjnagel (approved):
😭 that this is required, but LGTM

PR #510, opened 2023-08-18
chore: remove dead variables from DUBBD

Racer159:
This removes some dead variables from the DUBBD package.


mjnagel (approved):
Thanks @Racer159 !

MxNxPx (approved):
:ship: :it:

Issue #422, opened 2023-07-20
dubbd patterns for extensibility


MxNxPx:
as a platform engineer consuming dubbd, it would be helpful to have examples of how to extend dubbd, dubbd-k3d, and dubbd-aws if there is a need to layer on features or perform overrides.  the expectation is that feedback would be provided via github issues (or even better PRs raised) to dubbd, but still allow the flexibility to extend until merged & released.  one option is to provide and publish an example of using the zarf skeleton and extending specific pieces (namely: overriding helm values, or adding a component).

MxNxPx:
initial progress was made towards making dubbd-aws work as a zarf skeleton via [this PR](https://github.com/defenseunicorns/uds-package-dubbd/pull/486)

but placing this as blocked for now as the uds cli work will be a better solution

PR #511, opened 2023-08-18
feat: additional dubbd helm values via zarf var

MxNxPx:
to address https://github.com/defenseunicorns/uds-package-dubbd/issues/422

limitation to this is that it only allows the user to modify dubbd package helm chart values but does not support adding new ones (since this is injected at package deploy time, not package create time). also, the bigbang extension builds out the full zarf yaml at package create time, so toggling off a helm chart via `enabled: false` does not work as the zarf.yaml is expecting it and wait to for the package to be ready (eventually timeout/fail).


aws/dubbd-aws/zarf.yaml
```diff
@@ -49,6 +49,13 @@ variables:
     description: "List of namespaces to be exempt from cluster policies (must contain zarf)"
     default: '[ "zarf", "calico-apiserver", "calico-system", "tigera-operator" ]'
 
+  - name: ADD_DUBBD_VALUES
+    default: deploy-add-dubbd-values.yaml
+    prompt: false
+    type: file
+    autoIndent: true
```
mjnagel, line 56: I'm not so convinced anymore on the value of having a separate "extra values" specific to AWS/k3d/rke2...if someone is deploying DUBBD AWS what is the benefit of having two spots to override values? They could leverage the one from the base DUBBD, even if deploying AWS and I don't think there's any extra benefit to having two values files?

defense-unicorns-distro/zarf.yaml
```diff
@@ -54,6 +54,11 @@ variables:
   - name: EPHEMERAL
     description: "Set whether the cluster should be considered ephemeral - if true all resources will be force destroyed on removal"
     default: "false"
+  - name: DUBBD_VALUES
+    default: deploy-dubbd-values.yaml
```
mjnagel, line 58: I don't have a great suggestion here but I wonder if we can find a default that won't throw a warning on deploy if it's not provided 🤔 

```
 WARNING  Refusing to load a non-text file for templating ###ZARF_VAR_DUBBD_VALUES###                                                                                                                                   
                                                                                                                                                                                                                        
 WARNING  Refusing to load a non-text file for templating ###ZARF_VAR_ADD_DUBBD_VALUES###                                                                                                                               
```

mjnagel (commented):
Overall I like it. Do you think we need docs on intended usage and what is/isn't supported? Thinking primarily of someone stumbling on this and trying to enable/disable a package which wouldn't work. *Maybe* also something to encourage less usage of this and more usage of issues to suggest default values for DUBBD?

I know we don't necessarily want this "advertised" for usage, but I think it would be good to document it with caution signs and bold warnings to keep us out of trouble 😆

aws/dubbd-aws/zarf.yaml
mjnagel: As discussed synchronously - this was added to allow overriding the specific values for AWS, k3d, etc (due to the values file ordering). Might be possible to share the same zarf var to accomplish this same purpose...

defense-unicorns-distro/zarf.yaml
MxNxPx: @mjnagel - good catch!
addressed by including a default file (like [this one](https://github.com/defenseunicorns/uds-package-dubbd/blob/dubbd-values/defense-unicorns-distro/deploy-dubbd-values.yaml)) in the repo for each dubbd package and no longer receive that warning

mjnagel (approved):
:shipit:

aws/dubbd-aws/zarf.yaml
MxNxPx: @mjnagel - updated to use the same zarf var regardless of the package which will write to both the dubbd base values override file and the dubbd values override file for each package ([dubbd-aws](https://github.com/defenseunicorns/uds-package-dubbd/blob/dubbd-values/aws/dubbd-aws/zarf.yaml#L52), [rke2](https://github.com/defenseunicorns/uds-package-dubbd/blob/dubbd-values/rke2/zarf.yaml#L29), [k3d](https://github.com/defenseunicorns/uds-package-dubbd/blob/dubbd-values/k3d/zarf.yaml#L26)).  although redundant, it doesn't harm anything, and it simplifies it to a single zarf var that has the same impact of being the last values file with the highest precedence.

MxNxPx:
> docs on intended usage and what is/isn't supported? Thinking primarily of someone stumbling on this and trying to enable/disable a package which wouldn't work (or switch

added a description in the zarf.yaml files

PR #517, opened 2023-08-22
feat: add more preflight checks

noahpb:
addresses concerns for dubbd in https://github.com/defenseunicorns/uds-cli/issues/4

### DIFF
--- a/defense-unicorns-distro/scripts/preflight.sh
+++ b/defense-unicorns-distro/scripts/preflight.sh
@@ -1,14 +1,14 @@
 #!/bin/bash
 
-# Run your command and get the output
-output="###ZARF_VAR_OUTPUT###"
+# Retrieve output from test commands as zarf variables
+cluster_version="###ZARF_VAR_CLUSTER_VERSION###"
 has_kubeconfig="###ZARF_VAR_HAS_KUBECONFIG###"
 is_healthy="###ZARF_VAR_IS_HEALTHY###"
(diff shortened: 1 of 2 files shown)


defense-unicorns-distro/scripts/preflight.sh
```diff
@@ -15,9 +19,33 @@ supported_minor_version="26"
...
   echo "Server version is not supported"
-  echo "Please update your cluster to a minimum Kubernetes version of ${supported_major_version}.${server_minor_version}"
+  echo "Please update your cluster to a minimum Kubernetes version of ${supported_major_version}.${supported_minor_version}"
   exit 1
 fi
+# system has a kubecontext set
+if [ "$has_kubeconfig" == '""' ]; then 
+  echo "No kubecontext has been set. Please ensure that a valid context is configured before proceeding."
+  exit 1
+fi
```
mjnagel, line 31: Does it make sense to reverse the order of these two checks? And maybe the health one? Confirm we have a kubeconfig (and maybe confirm cluster health) before checking version?

MxNxPx (approved):
:ship: :it:

MxNxPx (changes requested):
looking good after you make some of the changes we discussed where the commands run in the component before action always return a 0 exit status and is handled with logic in the script after the fact.

MxNxPx (dismissed):

mjnagel (approved):
:shipit:

PR #554, opened 2023-09-06
fix: temp fix to secret used by kiali for grafana

MxNxPx:


zachariahmiller (dismissed):
🚀

defense-unicorns-distro/zarf.yaml
```diff
@@ -326,3 +326,21 @@ components:
...
+          - cmd: |
+              ./zarf tools kubectl apply -f kiali-grafana-temp-fix-credentials.yaml
+              ./zarf tools kubectl -n kiali delete pods -l app.kubernetes.io/name=kiali --ignore-not-found --timeout=2m
+            dir: run/manifests
+            description: "Create credentials for integration with Grafana & cycle Kiali pod(s)"
+      onRemove:
+        after:
+          - cmd: |
+              ./zarf tools kubectl delete -f kiali-grafana-temp-fix-credentials.yaml
+            dir: run/manifests
```
mjnagel, line 341: nit: this would require the file to still be available - instead we could delete by name so that this doesn't require the file.

mjnagel (approved):

defense-unicorns-distro/zarf.yaml
MxNxPx: @mjnagel - good point, updated to target the secret name

mjnagel (approved):
:shipit:

PR #557, opened 2023-09-08
feat: Add COMPLIANCE.md

CloudBeard:
Created COMPLIANCE.md to show high level what % of controls DUBBD meets for NIST 800-53 High, Moderate, and Low.

The images are created by a github action that uses the Javascript code here: https://github.com/defenseunicorns/compliance-md-generator/blob/main/generatePieChartSvg.js which scraps the DUBBD OSCAL file for the data.

Should re-run and update anytime OSCAL is updated, keeping the COMPLIANCE.md updated.


COMLIANCE.md
```diff
@@ -0,0 +1,256 @@
+# Defense Unicorns Big Bang Distro (DUBBD) Compliance
```
mjnagel, line 1: File name is missing `p` in compliance 🙃

COMLIANCE.md
```diff
@@ -0,0 +1,256 @@
...
+<td>
+
+## NIST 800-53 High
+
+<img src="./compliance-images/NistHighComparison.svg" alt="NIST High"/>
+
+</td>
+<td>
+
+## NIST 800-53 Moderate
```
mjnagel, line 19: Design thing and not sure what the fix is, but the moderate control image is slightly larger/not aligned with the other two:

Screenshot 2023-09-08 at 10 27 54 AM

MxNxPx: This may be too much work for little benefit...
How about switching to bar charts instead of [pie charts](https://www.ataccama.com/blog/why-pie-charts-are-evil#:~:text=Pie%20Charts%20can%E2%80%99t%20convey%20more%20information%20without%20becoming%20overly%20complex)?

If a good idea, I am also fine if this could be out of scope for the current PR and handled later.

CloudBeard: @mjnagel Im not 100% sure whats causing that since all 3 are the same size. Maybe if I shrink them all by a small margin.

@MxNxPx If we can roll with this I can start working on a branch to swap to bar charts. [D3](https://observablehq.com/@d3/gallery?utm_source=d3js-org&utm_medium=nav&utm_campaign=try-observable) has a ton of cool charts. I don't think creating more types of visual is wasted work as we could re-use that code when it comes time to take the visualization stuff to the next level.

MxNxPx: @CloudBeard - if @mjnagel doesn't object, i am fine with merging this as-is and perhaps have a follow up PR to work on resizing the images if possible.

MxNxPx (approved):
nice work @CloudBeard -- :+1:

PR #558, opened 2023-09-08
feat: default cert-manager

TristanHoladay:


k3d/deploy-cert-resources.yaml
```diff
@@ -0,0 +1,66 @@
+# apiVersion: cert-manager.io/v1
```
mjnagel, line 1: I don't know if this has to be a "now" thing but I wonder if there's ways we can abstract this from our end users. i.e. you pass in your issuer, some private token/access, and we create the appropriate issuer/certificates.

It's probably a larger task to approach properly, especially when trying to consider the different types of issuers, maybe you couldn't abstract enough common fields to make it worthwhile? But I think as an ideal goal that's what I would love to see - the ability to not worry about creating these specific CRs, but instead just passing in some config that is used to make them for me.

defense-unicorns-distro/cert-manager/zarf.yaml
```diff
@@ -0,0 +1,69 @@
+kind: ZarfPackageConfig
+metadata:
+  name: cert-manager
+  description: "Certificate manager for DUBBD"
+  # x-release-please-start-version
+  version: "1.12.4"
```
mjnagel, line 6: I think this will need to be the same as the current DUBBD version for release please to version it correctly. Would also need to add to https://github.com/defenseunicorns/uds-package-dubbd/blob/main/release-please-config.json#L13

defense-unicorns-distro/cert-manager/zarf.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - "registry1.dso.mil/ironbank/jetstack/cert-manager-webhook:v1.12.4"
+      - "registry1.dso.mil/ironbank/jetstack/cert-manager-controller:v1.12.4"
+      - "registry1.dso.mil/ironbank/jetstack/cert-manager-ctl:v1.12.4"
+
+  - name: cert-manager-custom-resources
+    required: true
+    manifests:
+      - name: cert-manager-custom-resources
+        files:
+          - ../../values/cert-manager-resources.yaml
```
mjnagel, line 50: I think to make sure this doesn't throw warnings on deploy, when not providing a `CERT_RESOURCES` file, you have to ensure that `deploy-cert-resources.yaml` is copied in as a file as well. Otherwise including those mostly empty files doesn't provide a ton of value.

Maybe a broader question - was there a discussion on whether we want cert-manager always on vs optional? If I don't want to use it I guess it would just deploy with cert-manager CRs and not be problematic...but it would be taking additional resources.

mjnagel (commented):
Overall don't see any glaring issues - just a few questions on this.

defense-unicorns-distro/zarf.yaml
```diff
@@ -377,3 +390,14 @@ components:
...
+    required: true
+    actions:
+      onDeploy:
+        after:
+          - cmd: ./zarf tools kubectl apply -f cert-manager-service-monitor.yaml
+            dir: run/manifests
+            description: "create service monitor for cert-manager after DUBBD has deployed"
+    files:
+      - source: manifests/cert-manager-service-monitor.yaml
+        target: run/manifests/cert-manager-service-monitor.yaml
```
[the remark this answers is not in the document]
TristanHoladay: I'd tried switching this yesterday to use `manifests` and got helm install errors. My thought was it's because of the labels / annotations already on the ServiceMonitor definition conflicting with Helm somehow (although one of the labels was 'managed-by: Helm' ...). I tried removing things but still was getting errors, so switched it back. I'll try again to see if I can get it working.

k3d/deploy-cert-resources.yaml
TristanHoladay: Agreed, that would be ideal. I think that will just take some time researching all the different issuer types (Self-Signed, CA, external (AWS, GCP, etc...), Venafi, Vault, ACME) to see if there's enough commonality to make the abstraction doable.

defense-unicorns-distro/cert-manager/zarf.yaml
TristanHoladay: changed version to current DUBBD 0.9.1 and added cert-manager path to release-please-config.json

k3d/deploy-cert-resources.yaml
mjnagel: Going to resolve this thread - @MxNxPx for awareness I think this would be a great next step with cert manager.

MxNxPx:
closing in favor of using [this](https://github.com/defenseunicorns/uds-package-cert-manager) in a uds bundle

PR #567, opened 2023-09-13
feat: allow specifying istio admin domain & certs

MxNxPx:
This feature would enable optional use of zarf vars to set the admin domain and admin cert/key for admin istio ingress gateway.  if not provided, it would use the same domain and cert/key for tenant and admin istio ingress gateways.


RothAndrew (commented):
Nothing jumps out at me as a red flag. You may want to configure things so that Renovate updates the new pinned version in .github/workflows/create-compliance-md-images.yaml

MxNxPx:
> Nothing jumps out at me as a red flag. You may want to configure things so that Renovate updates the new pinned version in .github/workflows/create-compliance-md-images.yaml

Thanks @RothAndrew - if OK, was planning on addressing that in a follow-up PR.

defense-unicorns-distro/zarf.yaml
```diff
@@ -13,14 +13,24 @@ metadata:
...
   - name: CERT_FILE
     default: bigbang.dev.cert
-    description: "This file contains the cert for the ingress gateways"
+    description: "This file contains the cert for the tenant ingress gateway"
     prompt: false
   - name: KEY_FILE
     default: bigbang.dev.key
-    description: "This file contains the key for the ingress gateways"
+    description: "This file contains the key for the tenant ingress gateway"
+    prompt: false
```
mjnagel, line 25: Bit of a nit but technically these all apply to both tenant and admin IF the admin ones are not specified, correct? Maybe could update descriptions to indicate that functionality?

Alternatively could rename these to `TENANT_*` and only use them for tenant things but I like the benefit of only having to specify one so not sure...

mjnagel (approved):

defense-unicorns-distro/zarf.yaml
MxNxPx: good point @mjnagel - updated the descriptions to be more accurate now.

mjnagel (approved):
:shipit: LGTM

PR #575, opened 2023-09-26
fix: imports for rke2

mjnagel:


MxNxPx (approved):
thanks for including the fix for the [TF issue](https://github.com/hashicorp/terraform-provider-aws/issues/33588)

Issue #574, opened 2023-09-23
Preflight k8s version check using deprecated cli arg


mjnagel:
The `--short` arg on `kubectl version` is deprecated. As of the next zarf version this command will no longer work due to zarf updating to a kubectl version that drops support for the short flag.

This should be updated to a new command working with the latest kubectl.

mjnagel:
https://github.com/defenseunicorns/uds-package-dubbd/blob/main/defense-unicorns-distro/zarf.yaml#L117

This is the specific line that will be problematic soon due to the deprecation, although depending on what format we get for that zarf var we may need to edit the preflight itself as well.

MxNxPx:
@mjnagel - wondering if it would be better to switch to something a bit more predictable across historical versions of `kubectl`.

maybe we could do: `zarf tools kubectl cluster-info dump -o jsonpath='{.items[0].status.nodeInfo.kubeletVersion}' 2>/dev/null || true`

mjnagel:
I don't think this is a hard blocker, just something to note...BSD grep ships with macos by default and does not support `-P` I believe. @rjferguson21 and I came up with something similar to workaround this issue temporarily but it didn't work on my machine until I installed GNU grep.

There's similar issues with BSD vs GNU sed options, and imo its OK to just say you must have GNU sed/grep installed. If there were an easy alternative to using `-P` that might be nice though - I'm not sure how much of a pain it is to fully rip out/replace the BSD tools (I suspect a bashrc/zshrc alias would not be enough to make zarf use the correct one).

PR #577, opened 2023-09-27
fix: update preflight to remove deprecated arg

MxNxPx:
resolves https://github.com/defenseunicorns/uds-package-dubbd/issues/574


mjnagel (approved):
Tested on 0.29.2 and a build of zarf from the `main` branch and worked well for both!

PR #586, opened 2023-10-03
fix: bump flux resources requests & limits via patch

MxNxPx:


aws/patches/flux-overrides-source-controller.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: apps/v1
```
mjnagel, line 1: Is there a reason this is in both spots (aws and root of repo)?

mjnagel: Er, maybe I just remembered the answer - I think the imports don't properly adjust filepath for this (aka zarf bug)?

MxNxPx: @mjnagel - you are correct, this is a zarf bug (at the moment).  i'm fine with waiting on the fix from zarf so that we don't have to have a duplicate dir/file.  current behavior if you don't have the extra copy is this error on build:

```console
     ERROR:  Failed to create package: unable to process bigbang extension: unable to build kustomization:
             trouble configuring builtin PatchTransformer with config: ` path:
             /defenseunicorns/uds-package-dubbd/aws/patches/flux-overrides-source-controller.yaml
             `: open
             /defenseunicorns/uds-package-dubbd/aws/patches/flux-overrides-source-controller.yaml:
             no such file or directory
```

mjnagel: I think I'm okay with either option - I was initially thinking the bug could cause issues with skeleton and/or DUBBD in a UDS bundle but I think it might be okay in either case? Not as confident in the skeleton assessment but for bundles the zarf package should be pre-built at that point so it shouldn't cause issues.

mikevanhemert (approved):

mjnagel (approved):

aws/patches/flux-overrides-source-controller.yaml
MxNxPx: waited until the zarf v0.30.1 release and removed the aws/patches directory as the bug is fixed.

PR #587, opened 2023-10-03
feat: rm external-dns from zarf pkg

MxNxPx:
Removing external-dns from DUBBD in favor of using https://github.com/defenseunicorns/uds-package-external-dns in a UDS bundle.


mjnagel (approved):
:shipit:

Issue #549, opened 2023-09-05
Default Big Bang Kyverno AllowedRegistries policy blocks deployments


RothAndrew:
which causes helm deployments like the one in [this package](https://github.com/defenseunicorns/uds-sso/issues/25) to fail.

After some discussions with the Zarf team and @MxNxPx it sounds like the best course of action may be to have DUBBD disable Big Bang's kyverno policy and use a different one that only looks at pods.

Issue #568, opened 2023-09-14
document advanced configuration options


MxNxPx:
as a platform engineer deploying dubbd, it is helpful that it can be deployed with minimal configuration options but have documentation of some advanced options, so that (if needed) these additional configuration options can be looked up.

PR #599, opened 2023-10-17
fix: update kyverno image reg policy for pods only

MxNxPx:
this change is to resolve https://github.com/defenseunicorns/uds-package-dubbd/issues/549


mikevanhemert (dismissed):
Looks good. Are there docs that should be updated to account for the new Zarf var? If not, should we write some?

@oates 👀

MxNxPx:
> Looks good. Are there docs that should be updated to account for the new Zarf var? If not, should we write some?
> 
> @oates 👀

thanks @mikevanhemert for the feedback.  in this case the change doesn't introduce a new var, but leverages it a bit differently under the hood.  if agreeable, there is [this github issue](https://github.com/defenseunicorns/uds-package-dubbd/issues/568) that will be worked in a later PR to document some of these "advanced configs" like the registry list zarf var.

k3d/values/local-dev.yaml
```diff
@@ -20,17 +19,31 @@ kyvernoPolicies:
...
+              match:
+                any:
+                  - resources:
+                      kinds:
+                        - Pod
+              validate:
+                foreach:
+                  - list: request.object.spec.template.spec.[ephemeralContainers, initContainers, containers][]
+                    pattern:
+                      image: registry1.dso.mil* | registry.dso.mil* | 127.0.0.1* | ###ZARF_VAR_APPROVED_REGISTRIES###
```
mjnagel, line 48: I'm trying to think this through - if we are only looking at the pod level now then do we want/need reg1/reg.dso.mil anymore? I would think we only need the zarf registry at this point, plus the option for extras to pass in on the off chance someone is not using zarf for a deployment?

k3d/values/local-dev.yaml
```diff
@@ -20,17 +19,31 @@ kyvernoPolicies:
...
+    additionalPolicies:
+      restrict-pod-image-registries:
+        spec:
+          rules:
+            - name: validate-registries
+              exclude:
+                any:
+                  - resources:
+                      namespaces:
+                        - metallb-system
```
mjnagel, line 38: Not saying we have to do this now, but since metallb is in Ironbank we could drop this override and document using the IB images (either via [zarf](https://github.com/defenseunicorns/uds-package-metallb) or via [kustomization](https://github.com/defenseunicorns/uds-package-metallb/blob/main/kustomizations/kustomization.yaml) directly). And I guess along the lines of my other comment maybe this isn't even needed if you deploy metallb with zarf regardless?

k3d/values/local-dev.yaml
MxNxPx: @mjnagel - good questions.  i was assuming we need to support resources deployed without zarf and possibly from a another registry that should be allowed (ex: an internal nexus/artifactory/harbor).  i think it is reasonable to remove the hardcoded reg1/reg.dso.mil from the list and require that to be provided via the zarf var.

k3d/values/local-dev.yaml
MxNxPx: i would agree that we should make this change, although i prefer to implement later if that is ok.

mjnagel: An alternative would be to default the var for k3d to include `quay.io/metallb/*`. Don't want to get too wrapped on this since it was this way before, just thinking of better maintenance options so we don't have to duplicate the policy itself.

values/kyverno-policies.yaml
```diff
@@ -4,11 +4,41 @@ kyvernoPolicies:
...
+          rules:
+            - name: validate-registries
+              match:
+                any:
+                  - resources:
+                      kinds:
+                        - Pod
+              validate:
+                foreach:
+                  - list: request.object.spec.template.spec.[ephemeralContainers, initContainers, containers][]
```
mjnagel, line 38: Just realized (I think) this is looking at the wrong path? Since we are the pod level it would just be `request.object.spec.[ephemeralContainers, initContainers, containers][]` I think?

MxNxPx: fixed the two references where this was incorrectly pointing to the deployment path instead of the pod

mjnagel (approved):
LGTM :shipit:

Issue #602, opened 2023-10-20
Unable to run `zarf init` after installing DUBBD v0.11.1+


RothAndrew:
On a cluster with DUBBD v0.11.1+ installed, `zarf init` fails due to the new kyverno policy. This effectively blocks users from upgrading the version of Zarf they are using until (1) a new version of DUBBD is released with a fix, and (2) they upgrade DUBBD to the new version that contains the fix.

![image (8)](https://github.com/defenseunicorns/uds-package-dubbd/assets/16000938/8c6cac00-0de7-4ffd-a148-ba0153a6fb61)

PR #607, opened 2023-10-26
fix: kyverno image registry policy exception

MxNxPx:
this change is meant to fix https://github.com/defenseunicorns/uds-package-dubbd/issues/602


mjnagel (approved):
Sorry for delay on the review, LGTM

Issue #635, opened 2023-11-29
Enable Velero by default in the RKE2 flavor


blancharda:
Add configuration variables for enabling Velero and configuring object storage within the RKE2 flavor of DUBBD.
Add a default backup configuration to snapshot cluster resources (manifests) and storage volumes.

blancharda:
Completed in #636

PR #636, opened 2023-11-29
Add Velero for RKE2

blancharda:
Enabling velero with a default config. Adding variables to configure object storage.
Addresses #635 

NOTE:
Adds an assumption that object storage is available for RKE2 deployments.


rke2/values/rke2-velero.yaml
```diff
@@ -0,0 +1,40 @@
+addons:
+  velero:
+    enabled: true
```
MxNxPx, line 3: @blancharda i assume we will want this to be the toggle point:
```yaml
addons:
  velero:
    enabled: "###ZARF_VAR_VELERO_ENABLED###"
```

rke2/zarf.yaml
```diff
@@ -32,6 +32,15 @@ variables:
     prompt: false
     type: file
     autoIndent: true
+  - name: VELERO_ENABLED
+    description: "Toggles velero on and off"
+    default: "true"
```
MxNxPx, line 37: thinking we may want to conditionally set this value, something like: https://github.com/defenseunicorns/uds-package-external-dns/blob/37d8850c47189c09d9d914a7c5c4dbff287489e0/zarf.yaml#L53-L63

so default to false unless `VELERO_S3_BUCKET` is set.

MxNxPx (changes requested):

rke2/zarf.yaml
blancharda: I'm not sure I love the idea of components being toggled on (or off) dynamically.
Bigbang itself occasionally does this, and it can be quite confusing to debug.

If I've explicitly configured `VELERO_ENABLED = "false"` I don't think I would want an action in the Zarf package to override that at deploy time.

That said, I do like the idea of some kind of pre-flight check or warning though if  `VELERO_ENABLED = "true"` and the bucket info is unset 🤔

blancharda: I could also be talked into the default being false for now.. but I do think it is something we are going to want in almost every (if not all) environments eventually.

MxNxPx: @blancharda - good points!  i was thinking more about not impacting folks who might already be using the dubbd-rke2 package rather than the confusion this could cause on deployment.

thinking through this a bit more, if the default was false, then zarf wouldn't include it on package create time.  also if it was true and at deploy time it was changed to false, at package create time it will build out the zarf actions to wait on the helmrelease to be ready, and fail at deploy time.  :thinking:

blancharda: The zarf actions could be conditioned in a way to check for the component being enabled.. but the package create time setting is a real problem...

We may just need to enable it "permanently" by setting `enabled: true` 🤔
(Technically someone could still disable it at deploy time via custom bigbang values).

blancharda: Upon further discussion, I think the path forward here is to require object storage as a pre-requisite to deployment for the RKE2 distro, mirroring the behavior of the AWS distro.

In CI we can deploy S3 buckets (yes.. it's a little funny for the RKE2 "on prem" distro, but will allow for some re-use with the AWS package CI code).

For dev work and actual installs, users will be expected to provide a solution prior to the DUBBD install.

.github/test-infra/rke2/main.tf
```diff
@@ -87,3 +87,101 @@ resource "null_resource" "kubeconfig" {
     command     = "aws s3 cp ${module.rke2.kubeconfig_path} ~/.kube/config"
   }
 }
+
+###########################################################
+################ S3 Bucket(s) #################
```
blancharda, line 92: Could be moved to a local module to make creating the loki bucket easier

rke2/zarf.yaml
```diff
@@ -42,6 +55,17 @@ components:
...
+    required: true
+    actions:
+      onDeploy:
+        after:
+          - cmd: ./rke2-preflight.sh
+          - cmd: rm rke2-preflight.sh
+    files:
+      - source: scripts/rke2-preflight.sh
+        target: rke2-preflight.sh
+        executable: true
```
mjnagel, line 68: We don't have a preflight for this on the AWS DUBBD...but I'm not sure if that's because we forgot or because its unnecessary? If a var is not defaulted and prompt is not false it should prompt for it, correct? Or does that behavior change with `--confirm`?

blancharda: I just tested to double check..
When using `--confirm` the deployment will continue with empty strings `('')` in the variables, even when `prompt: true` is set in the variable declaration. Test values below if you want to confirm..

Zarf var declaration:
```
variables:
  - name: CHECKING
    prompt: true
    description: just checking..
    #default: intentionally empty
 ```

* Create the zarf pakcage: `zarf package create .`
* Deploy the package (no confirm): `zarf package deploy zarf*.zst`
  * Pauses and prompts the user as expected
* Deploy the package with confirm: `zarf package deploy zarf*.zst --confirm`
  * Deploys the package with output that indicates an empty variable
  ```
  ...
  variables:  deployment-specific values that are set on each package deployment

  - name: CHECKING  currently set to '' (default)
    description: just checking..
    prompt: true
  ...
  ```
  
  I think we should leave in the preflight if we want setting the vars to be enforced.

File: rke2/values/rke2-velero-s3.yaml [hunk 5]
```diff
@@ -0,0 +1,39 @@
...
+            excludedNamespaces:
+              - kube-system
+              - flux
+              - velero
+            ttl: "240h"
+      serviceAccount:
+        server:
+          create: true
+          annotations:
+            eks.amazonaws.com/role-arn: "###ZARF_VAR_VELERO_S3_AWS_ROLE_ARN###"
```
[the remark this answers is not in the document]
blancharda: Agreed, the var is already gone. Copy/pasta strikes again. This config isn't even using the key/secret 🙈 

I'll push an update.

rke2/values/rke2-velero-s3.yaml
```diff
@@ -0,0 +1,39 @@
...
+          - name: default
+            provider: aws
+            bucket: "###ZARF_VAR_VELERO_S3_BUCKET###"
+            config:
+              region: "###ZARF_VAR_VELERO_S3_AWS_REGION###"
+        volumeSnapshotLocation:
+          - name: default
+            provider: aws
+            config:
+              region: "###ZARF_VAR_VELERO_S3_AWS_REGION###"
```
mjnagel, line 18: Missed this earlier but these all have `S3` in the name - the vars in the zarf yaml and preflight check do not. I'm not sure things are being passed in properly?

blancharda: 🤦 
I really should have just started a new PR.

rke2/values/rke2-velero-s3.yaml
```diff
@@ -0,0 +1,51 @@
...
+            aws_secret_access_key=###ZARF_VAR_VELERO_BUCKET_KEY_SECRET###
+      configuration:
+        backupStorageLocation:
+          - name: default
+            provider: aws
+            bucket: "###ZARF_VAR_VELERO_BUCKET###"
+            config:
+              region: "###ZARF_VAR_VELERO_BUCKET_REGION###"
+              s3ForcePathStyle: true
+              s3Url: http://minio.minio.svc.cluster.local:9000
```
MxNxPx, line 23: @blancharda - wondering if we would want to consider making these zarf vars also? 

i am good with merging as-is as the dubbd override values can also be used to change these.

rke2/zarf.yaml
```diff
@@ -71,6 +95,7 @@ components:
         version: "###ZARF_PKG_VAR_BIGBANG_VERSION###"
         valuesFiles:
           - values/add-dubbd-values.yaml
+          - values/rke2-velero-s3.yaml
```
MxNxPx, line 98: would want this file to be above the `add-dubbd-values.yaml` so it can override anything set.

MxNxPx (changes requested):

rke2/zarf.yaml
blancharda: Interesting, I was thinking of that file as the base, and the other files as the overlays, but if the convention is the other way that's an easy enough swap :)

rke2/values/rke2-velero-s3.yaml
blancharda: Good catch!
That should definitely be exposed, otherwise anyone using this in practice (and not using minio) would have to completely replace the Velero config. 

I sort of expect that to be necessary anyways, but it would be good to variablize this out so that there was potential reuse of the "default" config at least..

PR #638, opened 2023-11-30
chore: refactor all .gitignore files into single top-level .gitignore

### LINKED ISSUE #265: Add the extracted flux executible to the .gitignore
Flux is extracted into [run/flux](https://github.com/defenseunicorns/uds-package-dubbd/blob/main/defense-unicorns-distro/scripts/extract_flux.sh#L31-L37) at package create time. We should make sure this is included in the `.gitignore` to prevent accidentally checking it in.

Alternatively, we could extract flux into `tmp` and update the scripts/actions accordingly.

TristanHoladay:
resolves #265


TristanHoladay:
Closing because there are some "gotchas" with merging all .gitignore into 1 top level file and DUBBD will be EOL fairly soon.

PR #639, opened 2023-11-30
chore: update k8s from 1.26 to 1.27

### LINKED ISSUE #582: Update test infra to k8s 1.27
k8s 1.26.x will reach EOL at the end of February. Preferably prior to that point we should update all test infra to 1.27.x:
- k3d
- eks
- rke2

I think it's best to stay on n-1 for our minor version but EKS does [now support 1.28](https://aws.amazon.com/blogs/containers/amazon-eks-now-supports-kubernetes-version-1-28/) (as well as RKE2/k3d) if we wanted to jump to 1.28.x version.

TristanHoladay:
resolves #582 

looks like RKE2 is using v1.27.6 based on https://github.com/defenseunicorns/uds-rke2-image-builder/blob/93fa053565db5bcfa4524084b61c4411c48ea271/packer/aws/variables.pkr.hcl#L20 which builds the ami image used by the DUBBD RKE2 test infra https://github.com/defenseunicorns/uds-package-dubbd/blob/d2d50ca5793a3d08900fadc71421562e56ea7ffe/.github/test-infra/rke2/variables.tf#L28


mjnagel (commented):
re:RKE2, we'll have to update the [ami ID](https://github.com/defenseunicorns/uds-package-dubbd/blob/main/.github/test-infra/rke2/variables.tf#L29) to be on a 1.27 version I think. The latest ubuntu based one should be `ami-012992f851fa58d5b` based on https://github.com/defenseunicorns/uds-rke2-image-builder/actions/runs/6856097457/job/18642565548 , but the TF might also need changes to work still (currently depends on aws-cli being installed which got removed previously).

Could also use the image builder test infra as a reference, although that's a bigger diff from what we have here - https://github.com/defenseunicorns/uds-rke2-image-builder/tree/main/.github/test-infra/rke2-cluster . Don't know if that's a thing to do in this same PR or a different one, don't want to overload the review/changes.

justin-o12 (approved):
LGTM

mjnagel (approved):
🥳

MxNxPx (approved):
:ship: :it:

PR #644, opened 2023-12-06
chore: refactor dubbd-k3d ci to use the uds task runner

TristanHoladay:


.github/actions/setup/action.yaml
```diff
@@ -0,0 +1,25 @@
...
+      shell: bash
+      run: curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | TAG=v5.6.0 bash
+
+    - name: Set up Homebrew
+      uses: Homebrew/actions/setup-homebrew@master
+
+    - name: Install UDS CLI
+      shell: bash
+      # renovate: datasource=github-tags depName=defenseunicorns/uds-cli versioning=semver
+      run: brew install defenseunicorns/tap/uds@0.4.0
```
mjnagel, line 25:
```suggestion
      run: brew install defenseunicorns/tap/uds@0.4.1
```

k3d/tasks/setup.yaml
```diff
@@ -0,0 +1,10 @@
+tasks:
+  - name: k3d-test-cluster
+    actions:
+      - description: "Create the K3d cluster"
+        # renovate: datasource=github-tags depName=defenseunicorns/uds-k3d versioning=semver
+        cmd: "zarf package deploy oci://defenseunicorns/uds-k3d:0.1.12-multi --confirm"
+
+      - description: "Initialize the cluster with Zarf"
+        # renovate: datasource=github-tags depName=defenseunicorns/init versioning=semver
+        cmd: "zarf package deploy oci://defenseunicorns/init:v0.31.1-${UDS_ARCH} --confirm --components=git-server"
```
mjnagel, line 10:
```suggestion
        cmd: "zarf package deploy oci://defenseunicorns/init:v0.31.3-${UDS_ARCH} --confirm --components=git-server"
```

.github/workflows/test-k3d-package.yaml
```diff
@@ -14,33 +14,16 @@ jobs:
...
+          username: ${{ secrets.IRON_BANK_ROBOT_USERNAME }}
+          password: ${{ secrets.IRON_BANK_ROBOT_PASSWORD }}
 
-      - name: Remove DUBBD from k3d cluster
-        if: always()
-        run: zarf package remove zarf-package-*.tar.zst --confirm
+      - name: Test k3d package
         working-directory: k3d
-        timeout-minutes: 60
+        run: uds run test-k3d-pkg
```
mjnagel, line 29: I don't think it actually matters since these are ephemeral runners, but by flattening to a single stage like this we do lose the always remove/destroy.

I wonder if its also useful to separate out steps to give more granular output blocks vs just a single one in the actions logs (I know this is something uds-core also does though).

TristanHoladay: yeah that's a good point. I was following the uds-core convetion when I wrote this. I was thinking too that, although it's nice to only see 1 stage that just runs everything, it is more of a cognitive load for someone to understand what `test-k3d-pkg` does. If we break out the steps in the CI, then it's very clear each task used for each stage.

MxNxPx (approved):
:ship: :it:

PR #647, opened 2023-12-13
ci: update renovate registry1 creds

MxNxPx:
registry1.dso.mil robot creds were expiring on 13dec23, replaced with one that expires on 09jun23


mjnagel (approved):
Love it!

PR #652, opened 2023-12-14
Refactor use new AWS org for CI

justin-o12:


MxNxPx (approved):
as discussed some of the jobs are expected to fail and a followup PR will clean this up.

lgtm!

PR #657, opened 2023-12-18
refactor: Remove workflows in old CI account.

justin-o12:


justin-o12:
The upgrade workflow is still failing, because we do not have a DUBBD release whose IaC (specifically for Loki and Velero) will handle a permissions boundary.

MxNxPx (approved):
:ship: :it:

PR #665, opened 2024-01-16
feat: add initial lula lint in OSCAL creation and lula lint action

CloudBeard:
Added initial Lula Lint Action.

Checkouts code, setups Lula, and Lints the OSCAL file to ensure it is a schema compliant file.

Only triggers on changes to the OSCAL file.

Added Lula Lint inside of OSCAL generation to lint files at time of creation. This covers scenarios where OSCAL is created in CI from Component Generator or out of line updates.

### DIFF
--- a/.github/workflows/create-uds-oscal-component.yaml
+++ b/.github/workflows/create-uds-oscal-component.yaml
@@ -13,17 +13,17 @@ jobs:
 
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


mjnagel (commented):
I think this in general LGTM - not completely following why the workflow is in two spots though, could you explain that portion? I might just be having trouble following the PR description.

CloudBeard:
@mjnagel the lula-lint action would catch any changes to the OSCAL that could happen outside of the component generator. Its not a necessity where its included in the in the component generator portion.

PR #685, opened 2024-02-13
ci: explicit permissions set for calls to workflows

MxNxPx:


rjferguson21 (approved):

.github/workflows/tag-and-release.yaml
```diff
@@ -28,6 +28,8 @@ jobs:
     outputs:
       header_info: ${{ steps.header-info.outputs.header_info }}
       deploy_info: ${{ steps.deploy-info.outputs.deploy_info }}
+    permissions:
+      contents: write
```
zachariahmiller: I dont think write is needed here.

zachariahmiller: thanks

zachariahmiller (approved):
🚢 🇮🇹

PR #736, opened 2024-07-15
ci: add oci match to renovate github tag updates

MxNxPx:


mjnagel (approved):
Tested with https://regex101.com/r/2bBiYy/1 - seems like it should do the trick!

PR #740, opened 2024-07-16
ci: configure uds-k3d to utilize k8s (k3s) version 1.29

MxNxPx:


mjnagel (approved):
LGTM - if its helpful I have a PR on uds-k3d right now that would make this the default once released - https://github.com/defenseunicorns/uds-k3d/pull/92

